PR #30, opened 2020-11-11
Fix the fix of the angular velocity

stefan750:


Starystars67:
Lets hope this one does it 👍

PR #32, opened 2020-11-14
Added feature and fixed bug

jojos38:
Added the "skip other people vehicles when switching" feature and fixed quite a big bug with ownMap


Starystars67:
Nice work

PR #43, opened 2020-12-06
Chat rewrite

jojos38:
Cleaning of the chat system code
Fixed " giving lua error (by replacing them with single quote)
Improved chat design
Improved chat performances
Fixed not having the chat on the UI causing lua errors


Starystars67 (approved):
Nice work. Clean too!

PR #79, opened 2021-01-22
Added rules and tos acceptance (not deeply tested)

jojos38:
Added rules and tos acceptance
I tested this few times, so far it seems to work, further testing might be a good idea


20dka:
can you post a screenshot?

PR #84, opened 2021-02-03
MpVehicleType, favorites, js code slimming

20dka:


Starystars67:
If this is still good to merge then I would like to merge it into the new development branch which has been created in prep for moving to reimplement EA and Dev builds!

20dka:
yep its stable as far as i can see

PR #90, opened 2021-02-19
Fixed "ghost driving" other people's cars

prestonelam2003:
I reworked how input is synced on the client. The throttle,brake,clutch and handbrake are all direct input, the steering is smoothed so that watching other people drive is bearable, I can smooth the rest fairly easily if needed, I also tried to comment most of the stuff propperly


prestonelam2003:
need to fix some conflicts because github is a pain

PR #99, opened 2021-03-23
README Rewrite

lionkor:
Hey guys, 

I noticed the README was **super** out of date. I rewrote almost the entirety of it, mostly linking instead of copy-pasting the guide(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,32 @@
 # BeamMP
-A LUA implementation at attempting to bring local network multiplayer to BeamNG.drive
+A Lua mod bringing online multiplayer to [BeamNG.drive](https://beamng.com)
 
 # Contents
- - Installation
- - How does this work?
- - Network Codes
- - Progression tracking.
- - Troubleshooting - connecting to and playing on servers
- - Discord link
+ - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


lionkor:
Look at [this](https://github.com/BeamMP/BeamMP/tree/3d6086a7533e7fe47829a9f702acd35bb153753c) for a nice preview

Starystars67:
Well hot damn!
Looking pretty fine! One suggestion that might be worth having would be maybe have a section about the past of the project?
Just an idea that's all. Otherwise I am goo with a merge so I will approve now either way.

Starystars67 (approved):

lionkor:
Glad you like it! The past of the project sounds like a fun thing to write out, a short little history - but I'm likely the wrong person to write that. I think it should be a separate page on the website, maybe?

Starystars67:
Fair enough there. Do you think we should mention anything about network codes or leave that out / something for the wiki?

lionkor:
The network codes are quite arbitrary and the data format is also important, so we're better off writing a full guide on the wiki.
I'm all for Readme's that link instead of being *the* place to get any info you can't get anywhere else :D

PR #114, opened 2021-04-05
MPVehicleGE: Move nametag rendering from onUpdate into onPreRender

lionkor:
This reduces potentially reduces the load on physics-frames.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -784,34 +784,6 @@ local function groundmarkerFollowPlayer(targetName, dontfollow)
 	end
 end
 
-local function onPreRender(dt)
-	local vehicle = be:getPlayerVehicle(0)
-	if not vehicle then return end
-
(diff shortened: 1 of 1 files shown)


Anonymous-275 (approved):

Starystars67 (approved):
Does

lionkor:
We didn't benchmark this change, but `onUpdate` would be called every physics frame, which nametag rendering and plents of string formatting should not be tied to. 

I assume that this change won't lead to a noticable FPS difference (and if at all, then only with distance + fading turned on), but I would take an educated *guess* and say that most of the BeamMP lua codebase lacks major operations *that can easily be optimized*, and that the time lua takes to do it's thing is likely due to a large set of small operations like this one.

In my experience, implementing lots of small optimizations like this will have a pretty good impact on performance and stability, and I indent to implement a good count more. :)

Starystars67:
Nice Work. I know @Anonymous-275 was discussing things like this last night too. Nice to see little things being done to help keep it all moving!

PR #123, opened 2021-05-20
Updates to MPElectricsVE and MPInputsVE

jake55520:
MPElectricsVE

optimizations
-	removed tostring from localCurrentGear because it would result 
	in it never being equal to data (which is a number), causing
	the loop to run constantly

-	Added loads of unnecessary spammy electrics to the
	disallowed list, thanks to Olrosse for helping me find them
	Should reduce network traffic by a lot with certain vehicles

Improvements
-	Shifting into M mode on a DCT at speed no longer only shifts
	into M1, but checks if it's in 'S' or 'D' mode first then upshifts
	into the according M gear, preventing overrevving
-	Improved and compacted overall gear code for automatic and DCT transmission
-	Added firing rockets support to the ME262 plane
	thanks to Olrosse for finding the trigger
-	DH Super bolide wings sync
-	ESC Mode syncing
-	ABS Mode sync (no more smoke screen from players using arcade ABS)
-	City bus door and suspension level syncing, thanks to 20dka <3
Fixes
-	Fixed typo in trans brake sync

MPInputsVE

-	Fixed inputs getting stuck when no input was present, likely a bodge but it seems to work without any issues
	Switched all inputs back to input events because setting through electrics has a whole load of problems
	like hamster wheel throttle not doing anything, brake lights not working on some cars, the clutch outright not working
	and steering not actuating rear steering
	Changed the filters on the inputs so they're way quicker and more responsive to actual player inputs

### DIFF
--- a/lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
+++ b/lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
@@ -467,12 +467,12 @@ local function applyElectrics(data)
 		end
 		-- Bus door syncing
 		if decodedData.dooropen ~= nil then
-			local doorsController = controller.getController('doors')
+			local doorsController = controller.getControllerSafe('doors')
 			if doorsController then
 				if decodedData.dooropen == 1 then
-					controller.getControllerSafe('doors').setBeamMin({'frontDoors', 'rearDoors'}) -- open doors
+					doorsController.setBeamMin({'frontDoors', 'rearDoors'}) -- open doors
 				else
-					controller.getControllerSafe('doors').setBeamMax({'frontDoors', 'rearDoors'}) -- close doors
+					doorsController.setBeamMax({'frontDoors', 'rearDoors'}) -- close doors
 				end
 			end
 		end	
--- a/lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
+++ b/lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
@@ -67,19 +67,14 @@ local function applyGear(data)
 		local index = tonumber(string.sub(data, 2, 3))
 		if not index then return end
 		local gearIndex = electrics.values.gearIndex
-		print("state from data: "..state)
-		print("index from data: ".. index)
-		print("gearIndex from electrics: "..gearIndex)
 		if state == 'M' then
 			if localGearMode ~= 'M' then
 				if localGearMode == 'S' or localGearMode == 'D' then
 					controller.mainController.shiftUp() -- this is so it doesn't go into M1 when switching into M modes at higher speed
 					localGearMode = 'M'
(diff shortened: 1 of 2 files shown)


lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
```diff
@@ -270,15 +458,64 @@ local function applyElectrics(data)
...
 			end
 		end
-
-		-- Anything else
+		-- Bus door syncing
+		if decodedData.dooropen ~= nil then
+			local doorsController = controller.getController('doors')
+			if doorsController then
+				if decodedData.dooropen == 1 then
+					controller.getControllerSafe('doors').setBeamMin({'frontDoors', 'rearDoors'}) -- open doors
```
20dka, line 473: you make use of the doorsController var you created instead of calling getControllerSafe again

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
```diff
@@ -270,15 +458,64 @@ local function applyElectrics(data)
 				end
 			end
 		end
-
+		-- Player crouch snycing
```
Starystars67, line 461: spelling mistake :P

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
```diff
@@ -270,15 +458,64 @@ local function applyElectrics(data)
...
 		if decodedData.isCrouching ~= nil then
 			local playerController = controller.getController('playerController')
 			if playerController then
 				playerController.crouch(decodedData.isCrouching and -1 or 1)
 			end
 		end
-
-		-- Anything else
+		-- Bus door syncing
```
Starystars67, line 468: Do doors and controls sync when a user interacts with them now? E.g. when one opens the tailgate of the D-Series does it open for all?

Starystars67 (changes requested):
Couple of minor things. Nice work @jake55520

Starystars67:
There is also a fair bit of debug printing left in there. Can those be cleaned up to use the log system of the game or removed. Thank you.

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
jake55520: Whoops, and somehow missed the debug prints too, all fixed now :)

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
jake55520: afaik that has always worked, but it has to be the owner of the vehicle that's interacting with the tailgate

Starystars67 (approved):

20dka (approved):

PR #124, opened 2021-05-29
Development

Starystars67:
Fix for Registration link.
Fix for logging issues causing crashes.
Improvement of session cleanup / mod removal, (kissmp reported issue.)

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -105,7 +105,7 @@ local function isLauncherConnected()
 	return launcherConnectionStatus == 2
 end
 local function login(identifiers)
-	log('M', l, 'Attempting login')
+	log('M', loggerPrefix, 'Attempting login')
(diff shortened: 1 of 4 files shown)


Anonymous-275 (approved):

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -105,7 +105,7 @@ local function isLauncherConnected()
 	return launcherConnectionStatus == 2
 end
 local function login(identifiers)
-	log('M', l, 'Attempting login')
+	log('M', loggerPrefix, 'Attempting login')
```
20dka, line 108: We could change this to something more descriptive in the future

20dka (approved):

lua/ge/extensions/MPCoreNetwork.lua
Starystars67: indeed in the future we could but for now I see no issue with it unless you have an idea?

PR #125, opened 2021-06-23
[0.23 Fixes] + New Fixes for LineLock, Various other things + Couple of UI bits

Starystars67:
As the title says, this is a collection of work merging some of Jojos final work, Batman and co's debugging and fixes and some of my work too.

This update REQUIRES work to the launcher. Come please speak to me regarding this. The changes needed are the same ones that Jojos had spoken to Anonymous about.

Also includes the fix for KissMP to be run alongside BeamMP for those who use both.


lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
```diff
@@ -53,71 +55,111 @@ local function applyGear(data)
...
-		localGearMode = state
-		localCurrentGear = data
-
 	-- We use the same thing as automatic for all the first gears, and we use
 	-- the same type of shifting as sequential for M gears.
-	elseif gearboxType == "dctGearbox" or gearboxType == "cvtGearbox" then
+	elseif gearboxType == "dctGearbox" or gearboxType == "cvtGearbox" or gearboxType == "automaticGearbox" or gearboxType == "electricMotor" then
 		local state = string.sub(data, 1, 1)
+		local index = tonumber(string.sub(data, 2, 3))
+		if not index then return end
```
Olrosse: if not index then return end wasn't supposed to be there and prevents the automatic code from running if it's not in a state like M or S

lets see if i can work out github this time, don't know if i actually managed to post the other one earlier hehe

Anonymous-275 (approved):

stefan750 (approved):

PR #127, opened 2021-06-24
Fix for coupling code

Olrosse:
Couplers state was sending incorrect information and the detach code never actually detached the coupler, there are still one scenario i've found where it fails to attach and throws a failed to export electrics error, but its not fatal and i can replicated it in SP with no MP lua active so that appears to be a game issue


Starystars67 (approved):
Nice catch!

PR #134, opened 2021-07-08
Fix for broken Scenarios in SP

StanleyDudek:
Scenarios in singleplayer are broken while MP is in unpacked without this; this also lets things like uj_base's MOTD plugin work again

### DIFF
--- a/ui/entrypoints/main/main.js
+++ b/ui/entrypoints/main/main.js
@@ -1157,6 +1157,27 @@ angular.module('beamng.stuff')
   engineChangeData: null
 })
 
+.service('translateService', ['$translate', function($translate){
+  contextTranslate = function(val) {
+    if(typeof val == "string") {
+      return $translate.instant(val)
+    } else {
+      return $translate.instant(val.txt, val.context);
+    }
+  }
+  multiContextTranslate = function(val) {
+    let description = ""
+    for (var i = 0; i < val.length; i++) {
+      description = description + contextTranslate(val[i])
+    }
+    return description
+  }
+  return {
+    contextTranslate: contextTranslate,
+    multiContextTranslate: multiContextTranslate
+  }
+}])
+
 .service('gamepadNav', ['logger', '$rootScope', '$state', 'SpatialNavigation',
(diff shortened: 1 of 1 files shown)


ui/entrypoints/main/main.js
```diff
@@ -1157,6 +1157,27 @@ angular.module('beamng.stuff')
...
 
+.service('translateService', ['$translate', function($translate){
+  contextTranslate = function(val) {
+    if(typeof val == "string") {
+      return $translate.instant(val)
+    } else {
+      return $translate.instant(val.txt, val.context);
+    }
+  }
+  multiContextTranslate = function(val) {
```
legraphista, line 1168: @StanleyDudek `contextTranslate` and `multiContextTranslate` are declared globally here. Is this intended?

Starystars67: @legraphista by the looks of it @StanleyDudek has just followed the same lines as to how the original developers had written the file. Though please correct me if I am wrong @StanleyDudek ?

legraphista: @Starystars67 `contextTranslate`/`multiContextTranslate` are globally defined since they don't have `var`/`const`/`let` to anchor them to the scope. Other impl I've seen just defined functions within the scope 
`a = function(){}` is not the same as `function a(){}`

Starystars67 (approved):
@20dka @jake55520 Could you test this please before we merge it in. It does look all good from reading it but I am unable to test to confirm it is working code for the game. 

Once done so please review and approve. I already have done so. This way it can be merged as soon as it has been tested.

Starystars67:
I am testing this in my local dev build now.
(As best I can given I have major edits locally).

Starystars67:
Well game still runs and some of the errors in the console are now resolved which is good. There is no crashing or noticeable issues. Therefore my approval stands just need someone to also approve to be merged.

stefan750 (approved):
I tested it a bit and can confirm this fixes scenarios not starting properly, everything else seems to still work fine as well.

Starystars67:
Thanks Stefan, Awesome as always.

PR #138, opened 2021-07-26
small fix for a JS error

StanleyDudek:
this fixes JS error in console when players join a server, where the message is supposed to show in the top left (ie. "guest0123456 joined the game!"), and therefore restores that message appearing, and I guess would ensure any other MP notifications that take place there are shown.


Starystars67 (approved):
Interesting, Nice catch. Is this a change in the way the game handles the ui_message function now?

lionkor (approved):
Looks good

Issue #136, opened 2021-07-08
[Bug] Not returning to Main Menu when quitting from servers with mods


StanleyDudek:
**Have you read our FAQ/Wiki/#before-you-ask ?**
Yes

**Describe the bug**
When clicking the "Quit" button on the "MP-SESSION" UI app while loaded into a server that has mods, the player is disconnected from the server but not automatically returned to the main menu.

**To Reproduce**
1. Connect to any server with any mod
2. After loading in, click on "Quit" at top of screen

**Expected behavior**
Immediate return to Main Menu upon clicking "Quit"

**Desktop (please complete the following information):**
 - Windows Version 10.0.19043 Build 19043

**Additional context**
Behavior has been present since BeamNG.drive update 0.23.0.0.12345
- Screen flickers, MP UI apps disappear, player is now effectively in single player freeroam on the loaded map.
- No return to Main Menu, after clicking "Quit" in the "MP-SESSION" UI app, user must click "Quit to Main Menu" in lower left of UI to return

Starystars67:
Ohh interesting. They did change how a load of the UI works when transitioning so maybe they changed something with this too.
Thank you for bringing this to our attention and I will look into it this evening unless anyone else gets it done before I get in.

Starystars67:
Hello, I can confirm that this is fixed as part of our next update.

Starystars67:
This issue should now be fixed as of the PR above. It will be pushed out in due course

PR #141, opened 2021-07-30
Fix for issue #136

Starystars67:
As the title says.


Olrosse:
doesn't appear to be fixed, at least not on my end, on vanilla servers it leaves no problem, but on servers with mods it only removes the cars and UI but never goes back to main menu

Starystars67:
> doesn't appear to be fixed, at least not on my end, on vanilla servers it leaves no problem, but on servers with mods it only removes the cars and UI but never goes back to main menu

Hmm odd, the code is no different for either.

PR #144, opened 2021-09-10
Chat app improvements (and preventing the doublemessage issue)

5TXdYWK:
This PR adds a few chat app improvements: 

image

- Smaller font size for higher information density (and, subjectively, a better look)
- Smaller, unobtrusive timestamp
- Alternating backgrounds for messages for improved readability
- Formatting support (server messages only!)

~This also includes a simple, if dirty, fix to the doublemessage issues.
In case the deregistering doesn't work, we have a secondary catch in the form of a variable that keeps track of the last message (timestamped), so that it's highly unlikely we'll see doublemessaging again.~

### DIFF
--- a/ui/modules/apps/BeamMP-Chat/app.js
+++ b/ui/modules/apps/BeamMP-Chat/app.js
@@ -14,7 +14,7 @@ app.directive('multiplayerchat', [function () {
 }]); 
 
 
-app.controller("Chat", ['$scope', '$rootScope', function ($scope, $rootScope) {
+app.controller("Chat", ['$scope', function ($scope) {
 	$scope.init = function() {
     // try to unregister existing chatMessage listener
     $rootScope.$$listeners.chatMessage = [] 
@@ -92,9 +92,8 @@ app.controller("Chat", ['$scope', '$rootScope', function ($scope, $rootScope) {
 		else setChatDirection("bottom");
 	}
 
-	$rootScope.$on('chatMessage', function (event, message) {
-    if (event.target === event.scope) {
-		addMessage(message);}
+	$scope.$on('chatMessage', function (event, message) {
+		addMessage(message);
 	});
 
 	$scope.chatSend = function() {
--- a/ui/modules/apps/BeamMP-Chat/app.js
+++ b/ui/modules/apps/BeamMP-Chat/app.js
@@ -92,8 +92,9 @@ app.controller("Chat", ['$scope', '$rootScope', function ($scope, $rootScope) {
 		else setChatDirection("bottom");
 	}
 
-	$scope.$on('chatMessage', function (event, message) {
-		addMessage(message);
+	$rootScope.$on('chatMessage', function (event, message) {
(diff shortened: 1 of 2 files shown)


ui/modules/apps/BeamMP-Chat/app.js
```diff
@@ -169,40 +174,132 @@ async function showChat() {
...
-	let node = document.createElement("li");
-	node.style.padding = "3px 0px 3px 0px";
-	node.style.backgroundColor = "rgba(0, 0, 0, 0.45)";
-	fadeNode(node);
-	let textNode = document.createTextNode(msg);
-	let chatlist = document.getElementById("chat-list");
-	node.appendChild(textNode);
-	chatlist.appendChild(node);
+  // make sure that the last message isn't a dupe
+  if (msg === lastReceivedMessage) {
```
lionkor, line 254: To me this doesn't seem right at all.

If you send two messages in the same second with the same content, that should still display as two messages.

lionkor: I understand what this should do, but it needs to never, ever block out a legit message, especially when someone is spamming it in <1 sec intervals.

Starystars67: > I understand what this should do, but it needs to never, ever block out a legit message, especially when someone is spamming it in <1 sec intervals.

I would argue that spamming like that should not be allowed anyway so in this example I would argue that there is no real issue. One issue I do see is if two players send the same message at the same time then only one of their messages will show which could be bad. Could we add a check that the username is different maybe? OR do a true timestamp comparison (Of the time sent)?

5TXdYWK: @lionkor I've removed the workaround!

Starystars67 (changes requested):
Firstly, Thank you for helping towards the project! Do you have any intentions of helping beyond this work now? If so I can add you to the contributors text channel in our discord.

Getting back to the PR, Looks great! Love the time stamp design and smaller text. I had too been thinking that the text felt rather on the large side. Text formatting is a nice touch too.

The only issue I have with this PR is indenting. Please can you just go over your work and ensure that your indenting is correct and consistent please. Once done I am happy to merge providing you are ready for it to be? Thanks again!

5TXdYWK:
> Firstly, Thank you for helping towards the project! Do you have any intentions of helping beyond this work now? If so I can add you to the contributors text channel in our discord.

Yeah, I'd like to help where I can.

> The only issue I have with this PR is indenting. Please can you just go over your work and ensure that your indenting is correct and consistent please. Once done I am happy to merge providing you are ready for it to be? Thanks again!

My editor made a mess of things - I think I got everything now. If you spot anything I missed do let me know!

Starystars67 (approved):

Starystars67:
Nice work. I will request the 2nd review from some of the guys and also add you to the discord channel!

Thank you again.

20dka (approved):

PR #145, opened 2021-09-14
Increase the launcher timeout value

DmitrijVC:
Connection to the launcher will be considered "lost" after 10 seconds instead of 2.

When spawning a car the game sometimes freezes for more than 2 seconds which results in lost connection to the launcher and also the server. If connection was lost other players will appear as stationary vehicles. 
This simple fix worked for me and other people. Although I am not a Lua programmer and didn't read too much source code of this mod, so that fix could be unstable, or break other logic.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -342,7 +342,7 @@ local function onUpdate(dt)
 			else send('Up') end -- Server heartbeat
 		end
 		-- Check the launcher connection
-		if launcherConnectionTimer > 2 then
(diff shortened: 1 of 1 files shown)


Starystars67 (approved):

Olrosse:
> 
> 
> I just confirmed with @Anonymous-275 that we tried this before and increasing this value only delays the issue.

just tried this with a guy ho disconnected every time he spawned, setting it to 10 has fixed it so far for him

Starystars67:
This issue should have been fixed as a part of Pull Request #148 . 
Thank you for taking the time towards trying to find a solution to the issue.

PR #151, opened 2021-10-12
Several fixes including a fix for desync after instabilities, camera glitch cause by queue, a fix for ignition aswell as hazard lights

Olrosse:
also fix for remote vehicle turning local, fix for unicycle instability loop and more, the individual commits should have most of the change-log

and added an experimental option to disable instability pausing, it is toggleable live in game and automatically reverts when leaving a server

i might PR a rewrite of the ignition code that is more robust later, it is just a bit messy right now so i will try to simplify it a bit first

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -218,7 +218,7 @@ local function applyVehEdit(serverID, data)
 			if configChanged then
 				--veh:setDynDataFieldbyName("autoEnterVehicle", 0, (be:getPlayerVehicle(0) and be:getPlayerVehicle(0):getID() == gameVehicleID) or false) -- this only works one way :(
 				veh:respawn(serialize(playerVehicle.config))
-			else
+			elseif vehicleConfig.paints then
 				log('I','applyVehEdit', "only color changed")
 				for k, v in pairs(vehicleConfig.paints) do
 					extensions.core_vehicle_manager.liveUpdateVehicleColors(gameVehicleID, veh, k, v)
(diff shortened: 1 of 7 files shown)


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -887,17 +887,17 @@ local function applyQueuedEvents()
...
 		UI.updateQueue(vehicleSpawnQueue or {}, vehicleEditQueue or {})
 	end
 
 	--if not vehicleEditQueue then return end
 	--dump(vehicleEditQueue)
 	for vehicleID, edit in pairs(vehicleEditQueue) do
 		log('I', "applyQueuedEvents", "Applying queued edit")
-		applyVehEdit(vehicleID, edit)
 		vehicleEditQueue[vehicleID] = nil
+		applyVehEdit(vehicleID, edit)
```
Starystars67, line 900: Why did you do the apply after removal from the array? Not complaining just trying to understand your theory.

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
```diff
@@ -462,8 +462,8 @@ local function applyElectrics(data)
 		end
 
 		-- Ignition syncing
-		if decodedData.ignition ~= nil then
-			if electrics.values.ignition ~= decodedData.ignition then
+		if decodedData.ignition ~= nil and electrics.values.ignition ~= nil then
```
Starystars67, line 465: Good find, Why not use `Number()` or similar? to convert it to a number rather than Boolean?

Starystars67: Might allow for finer control?

Starystars67 (commented):
Nice work, Just a couple of comments if you could please address them?

lua/vehicle/extensions/BeamMP/MPElectricsVE.lua
Olrosse: Do you happen to know why the data is received as a boolian in the first place when the electric number locally is a number?

Olrosse: Fair enough :) yeah i think it is good for mering, unless you have anything else to note?

Starystars67 (approved):

lionkor (approved):
Looks great, can't wait to try this!

20dka (approved):

PR #164, opened 2021-11-09
Various things ~ v4.1.3?

Starystars67:
I think this might be version 4.1.3 or 4.2.0? Not sure just yet so lets decide please. Maybe include some other bits too.


20dka:
> I think this might be version 4.1.3 or 4.2.0? Not sure just yet so lets decide please.
I vote for version 4.20

lionkor:
Is there a dependency between this and #211?

PR #165, opened 2021-11-10
Fix for reconnecting bug with launcher

lionkor:
This fixes a bug where, after briefly disconnecting, the launcher repeatedly tries to reconnect.

I don't know if this is the same issue as the issue we've been seeing for months, it might be.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -341,6 +341,7 @@ end
 
 
 local function onUpdate(dt)
+	launcherTimeout = 0
 	--print(launcherConnectionTimer)
 	--====================================================== DATA RECEIVE ======================================================
 	if launcherConnectionStatus > 0 then -- If player is connecting or connected


lionkor:
That entire piece of code is super suspect, `git blame` says that @jojos38 added this in 7028ba0f6b23550fd4e8cb2325a3b0ed9cfc8ae2, so maybe he knows why/what/how this is a thing here.

Starystars67:
Could have been a mistake? IIRC it was originally declared further up in the file? Either way nice find!

Starystars67 (approved):

lionkor:
The missing line for sure, I was more wondering about the entire logic in that block.

20dka (changes requested):
This effectively bypasses our "first connection check" system originally put in place so the BeamMP client mod can delete itself if it is being ran without the launcher present.
Instead we should disable the system properly, as it has been found to have been made obsolete.

lionkor:
> This effectively bypasses our "first connection check" system originally put in place so the BeamMP client mod can delete itself if it is being ran without the launcher present. Instead we should disable the system properly, as it has been found to have been made obsolete.

I didn't see that this effectively disables the entire piece of code, but I see that now. This should not be merged then and instead a proper fix should be implemented.

I agree the entire mod deletion is obsolete and can go.

lionkor:
I assume the inactivity means this was fixed already or isnt desired, closing

PR #208, opened 2022-02-16
Stop hiding distance when showDistance option not disabled; Allow 1500m distance fade

5TXdYWK:


Starystars67 (approved):

20dka (approved):

5TXdYWK:
@Starystars67 @20dka looks like this is approved, can we get it merged? 👀

PR #215, opened 2022-04-24
Fix nametags "shortening" to longer than they were previously

Mack29446:


Starystars67 (approved):
Looks alright to me.

Olrosse:
When i tested this a while ago i noticed that it shorts names that are the same amount of character after its shortened, could it be made so it only shortens if it ends up with less characters?

snepsnepsnep (approved):

PR #216, opened 2022-04-25
Fixed electrics de-sync if vehicle has no engine

Olrosse:
apparently "return end" isn't as local as i thought and it actually stops the entire function not just the ignition code it was in, whoops xD


snepsnepsnep (approved):

lionkor:
Can't test this right now, @20dka you please?

lionkor (approved):
very clean, let's push this tonight!

PR #219, opened 2022-06-07
Vehicle sync improvements

stefan750:
- Use stack instead of recursive function for finding connected vehicle parts (~3x faster)
- Also recalculate COG when vehicle is damaged, not just when beams break
- Added 0.2 second delay for recalculating COG and connected parts
- Fixed acceleration prediction being broken (thanks @Olrosse for pointing this one out!)
- Adjusted teleport rotation threshold to reduce cars getting stuck in objects

Changes are client side only and dont require any updates to Launcher or Server.

Did not cause any obvious issues during initial testing and should be ready to merge.

### DIFF
--- a/lua/vehicle/extensions/BeamMP/velocityVE.lua
+++ b/lua/vehicle/extensions/BeamMP/velocityVE.lua
@@ -7,86 +7,119 @@
 
 local M = {}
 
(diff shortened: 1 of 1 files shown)


Starystars67 (approved):
Looks good to me, Interested to get the next major update done finally.

snepsnepsnep (approved):

PR #222, opened 2022-06-16
Fix: table server list width greater than view on small display

Victortunes:
Simple modification, using flexboxes but does not respect the proportions of the columns

PR #233, opened 2022-09-13
Fix table server list width greater than view on small display (fixed branch)

lionkor:
Fixed version of #222 

closes #222

cherry-picked the commits from master to development


lionkor:
(it was about 5 commands to fix this)

Mack29446:
Replaced by #248

PR #248, opened 2022-09-23
Fix victortunes 2

Mack29446:
Fixed version of #233, which fixed #222 

Closes #222 

cherry-picked the commits from @lionkor's branch to development


Mack29446:
Just tested this - the Location, Players and PPS fields are way too large. 1440x900 monitor

PR #251, opened 2022-09-24
disable some nodes onInit

Anonymous-275:
- disable nodes using setNode
- add logic to vehicle type filter
- filter nodes that have wheel ID


Anonymous-275:
Things left to do:
- Make an option that turns it on or off

Starystars67:
Nice work, lets get the setting/option added before considering merging this though.

Anonymous-275:
> Nice work, lets get the setting/option added before considering merging this though.

yeah the thing is I need someone who has done just that before to do it

PR #253, opened 2022-09-25
Feature use simplified jbeams if available

20dka:


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -37,6 +37,17 @@ local roleToInfo = {
 	['MDEV']	= { backcolor = ColorI(194, 055, 055, 127), tag = " [BeamMP Dev]", shorttag = " [Dev]" }
 }
 
+local simplified_vehicles = {
+    etk800 = "simple_traffic_etk800",
+    etkc = "simple_traffic_etkc",
+    etki = "simple_traffic_etki",
+    legran = "simple_traffic_legran",
+    midsize = "simple_traffic_midsize",
+    pessima = "simple_traffic_pessima_parked",
```
Mack29446: Please change this to a non-parked version of the pessima

Mack29446 (changes requested):

Olrosse:
Unless i missed it it seems like the covet is missing from the list, i think that might be one of the cars with the name simplified instead of simple

20dka:
> Unless i missed it it seems like the covet is missing from the list, i think that might be one of the cars with the name simplified instead of simple

fixed!

Mack29446 (approved):
Thank you for making the relevant changes. I have tested this and it works.

snepsnepsnep (approved):

PR #254, opened 2022-09-25
IPC Communication - Launcher V3

Starystars67:
This code changes the connection from TCP/UDP to IPC which remove the network stack from processing and removes latency.


Starystars67:
Testing last night with ~5 people gave good stable results.

Starystars67:
Note that this PR requires Launcher V3 and if this gets pushed public the new launcher will need to be released with it.

Starystars67:
Probably worth scheduling this update to be released with the new launcher and update checking system from the backend.

20dka:
issues so far:
-ping seem to disappear after a while
-leaving a server shows the 'you have been kicked' prompt, sometimes does not actually let you leave
-mods remain after leaving a modded server

PR #255, opened 2022-09-26
Tweaks to coupler code

Olrosse:
Removed spam on spawns and on resets.

added checks to prevent it from sending double the necessary data.

Fixed a GE error caused by it trying to couple a car that doesn't exist.


Starystars67 (approved):
Looks good to me but I am unable to test the code right now.

If someone else can test this code that would be great. Thank you

snepsnepsnep (approved):

Mack29446 (approved):
I have tested this and it works well, thank you! :D

PR #257, opened 2022-09-29
Fix login json string not escaping properly

### LINKED ISSUE #139: [Bug] Unable to login when having the "^" character in your password
Yes

**Describe the bug**
When you try to login ingame with a password containing the special character "^". You'll get an the following error in the console:
Unexpected token ^ in JSON at position 62.
And then in game I get. Invalid answer from authentication servers, please try again later.

The password had been generated by a password manager.
It also contains other special characters like "{".

Is this maybe a security flaw?? I dont realy know. Not in my alley.

**To Reproduce**
1. Change your password to one containing the character or create a new account.
2. Try to login ingame with that account.
3. See error.

**Expected behavior**
That the server can process those characters just fine and let me login.

**Screenshots**
[Console error](https://cdn.discordapp.com/attachments/665263790673231882/867789755139620864/unknown.png)

 - Windows 10 Home 10.0.19042 build 19042

**Additional context**
None

snepsnepsnep:
json.stringify doesn't properly escape `' "` (and some other) characters which causes the login to fail.
Fixed by serializing the data to lua and then doing a jsonEncode from there.

Tested with a few different passwords which contain characters that previously caused issues and they all work as expected.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -108,8 +108,8 @@ local function isLauncherConnected()
 	return launcherConnectionStatus == 2
 end
 local function login(identifiers)
-	log('M', 'login', 'Attempting login')
-	send('N:'..identifiers)
+	log('M', 'login', 'Attempting login...')
+	send('N:'..jsonEncode(identifiers))
 end
 local function autoLogin()
 	send('Nc')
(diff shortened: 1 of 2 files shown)


lionkor:
Noob question but this doesnt stay around, right? Like, other mods cant read your credentials, right? (https://github.com/BeamMP/BeamMP/pull/257/files#diff-cb8a69ebbaa7c4a0117167c0080f2ce31e561db88232112b1e7fd63d3a293546R83)

Mack29446 (approved):
Looks good, thank you for fixing this annoying issue! I have not had the chance to test however

Olrosse (approved):

PR #258, opened 2022-10-03
stop mp gamestate from loading in freeroam

snepsnepsnep:
Currently we completely break the freeroam mode if BeamMP is loaded since https://github.com/BeamMP/BeamMP/pull/239
(MP UI is loaded instead of the freeroam one, and missions don't appear / big map mode doesn't work)
This addresses that issue.

### DIFF
--- a/lua/ge/extensions/multiplayer/multiplayer.lua
+++ b/lua/ge/extensions/multiplayer/multiplayer.lua
@@ -105,8 +105,6 @@ end
 
 local function onClientPostStartMission()
 	if MPCoreNetwork.isMPSession() then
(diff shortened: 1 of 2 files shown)


Mack29446 (dismissed):
Looks good, however I have not been able to test

Olrosse (dismissed):
Seems to be working :) tested freeroam with using the map, campaign, Light runner and Multiplayer

Mack29446 (approved):
Thank you for this. For reference, I still have not tested.

Olrosse (approved):

PR #259, opened 2022-10-04
fix onExtensionUnloaded() not being called on exit

### LINKED ISSUE #260: [Bug] onExtensionUnloaded() is not called on server exit.
Yes

**Describe the bug**
`onExtensionUnloaded()` is not called for loaded mods when exiting a server, presumably because they are deleted outright when cleaning up the mods.

**To Reproduce**
1. Join a server with mods that have extensions containing the `onExtensionUnloaded()` function, with something like a `log` or `print` inside so you can check the beamng.log to see if it has been called. 
2. Click on 'Quit'
3. Open beamng.log
4. See that `onExtensionUnloaded()` was never called
5. Also see that because it was not unloaded, extensions prints `extension unavailable: "modName" at location: "modName"`

**Expected behavior**
When exiting a server, any mods that have extensions containing an `onExtensionUnloaded()` function should have that function called, as some mods need to know they've been unloaded.

 - Win11 22H2

**Additional context**
I opened a pull request with a fix @ https://github.com/BeamMP/BeamMP/pull/259

StanleyDudek:
makes sure `onExtensionUnloaded()` is called for all loaded mod extensions EXCEPT for those from BeamMP and the base game.

### DIFF
--- a/lua/ge/extensions/MPModManager.lua
+++ b/lua/ge/extensions/MPModManager.lua
@@ -20,7 +20,7 @@ local cleanuptimer = 0
 
 
 --this should be every extension that might possibly be running, EXCEPT for the extensions loaded by mods on the server.
---putting them in this table to be used by extensions.unloadExcept() effectively means ONLY extensions loaded by mods on a server will have onExtensionUnloaded 
+--putting them in this table to be used by extensions.unloadExcept() effectively means ONLY extensions loaded by mods on a server will have onExtensionUnloaded() called
 local extensionExceptions = {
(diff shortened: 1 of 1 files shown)


Mack29446 (approved):
Thank you for this fix. I have tested this and it does not break the functionality of BeamMP, but I have not tested `onExtensionUnloaded()`

StanleyDudek:
As covered here: https://github.com/BeamMP/BeamMP/issues/260 , this is not necessary, `queueExtensionToLoad()` and `loadGameModeModules()` should be used in modScript.lua instead of `load()` and  `registerCoreModule()`. Thanks again, closing this PR now.

PR #261, opened 2022-10-07
fix guest login

snepsnepsnep:


Mack29446 (dismissed):
Thank you for this fix. Please note I have not tested this
 
 Update: I have now tested this, and it works.

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -109,7 +109,8 @@ local function isLauncherConnected()
 end
 local function login(identifiers)
 	log('M', 'login', 'Attempting login...')
-	send('N:'..jsonEncode(identifiers))
+	if not identifiers then identifiers = "" else identifiers = jsonEncode(identifiers) end
```
20dka: I would clarify that this chunk is only necessary for guests
```suggestion
	if not identifiers then identifiers = "" else identifiers = jsonEncode(identifiers) end -- guest login fix
```

Mack29446 (approved):
Thank you for the comment.

20dka (approved):

PR #263, opened 2022-10-09
add ui apps check and remove radial menu check

StanleyDudek:
I don't think the reason for checking if the radial menu and removing it exists anymore, at least in BeamNG.drive v0.26, and for some versions now where it is not a default base game UI app. (I think it used to break things?). The radial menu is baked into the in-game menu these days, even though it can still be added as a UI app, and neither of these seem to cause issues.

I have also added a way to check if a player has REMOVED any of the three BeamMP UI apps from the multiplayer UI apps layout and replace whatever is missing with its default position and size settings upon returning to the main menu after exiting a server, or whenever the MP mod is loaded.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -393,33 +393,66 @@ end
 
 -- ====================================== ENTRY POINT ======================================
 local function onExtensionLoaded()
-	-- removing the radial menu from the Multiplayer UI layout if it's present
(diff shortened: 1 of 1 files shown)


snepsnepsnep:
I don't think forcing the multiplayer apps is a good idea, as they're not required for MP to function and have been moved to use UI hooks so no errors are caused if they're missing.

As for the radial UI menu checking I have already removed that on my branch, it was only necessary for a brief period of time when BeamNG moved the radial menu from the freeroam layout to the escape menu.

StanleyDudek:
In that case, disregard :), I'm behind on things

PR #265, opened 2022-10-11
fix metallic paints error when spawning

snepsnepsnep:
Fixes error that prevents some vehicles and objects from spawning in some cases.


Olrosse (approved):

Mack29446 (approved):
Thank you for this. However, I have not tested this. Please ensure this has been tested before merging

20dka (approved):

PR #281, opened 2022-12-14
Fix traffic spawn crash when joining server

### LINKED ISSUE #280: [Bug] Game crashing when joining a server while AI traffic spawning is active
Yes/No

**Describe the bug**
The game crashes when joining a server while the AI traffic option "Start freeroam with traffic" is enabled

**To Reproduce**
- Enable "Start freeroam with traffic"
- Set "Max amount of vehicles" for AI traffic to more than the servers per-player car cap
- Join the server
- As soon as the game goes from loading the map to loading the car(s) the game crashes, since spawning cars that exceed the servers per-player car cap is suppressed by multiplayer

**Expected behavior**
Joining a server, regardless if "Start freeroam with traffic" is enabled or "Max amount of vehicles" exceeds per-player car cap, should not crash the game.

Either force-disable "Start freeroam with traffic" when joining a server, and/or limit "Max amount of vehicles" to the servers per-player car cap or force it to 0

**Screenshots**
image

snepsnepsnep:
If the server vehicle limit is below the amount of vehicles the traffic system tries to spawn the game will crash.
This stops the game from spawning traffic vehicles when joining a server.
Freeroam behavior is unaffected.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -27,6 +27,7 @@ local loggedIn = false
 -- event functions
 local onLauncherConnected = nop
 local runPostJoin = nop
+local originalFreeroamOnPlayerCameraReady
 
 local loadMods = false -- gets set to true when mods should get loaded
 --[[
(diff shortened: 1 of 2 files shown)


Olrosse (approved):
tested and it works on my end

20dka (approved):
we should look for a cleaner solution in the future, but this should be sufficient for now

PR #282, opened 2022-12-15
unload mod if game version is incompatible

### LINKED ISSUE #276: if GameVersion != SupportedVersion then abort BeamMP load
**A simple check on init of the lua mod that compares the game version against the supported game version.**

Whenever a mayor update of BeamNG drops _the mod becomes incompatible and usually breakes in a way where the players have no idea from where the issues are coming from_. By checking the game version the mod could abort his load and therefore save the players from these bugs.

By not doing this, as it is like know, _every BeamNG related place is flooded with support request_. That includes the BeamNG dc, BeamMP dc, reddit, the forum and what not. This puts more stress on these channels then it should be and it annoys lots of people.

_Players all over the place suggest to each other to rather remove the mod_ (**including BeamNG officially**) because of how "often" it breaks and some suggest the usage of KissMP. So the influence of the current situation is bad for BeamMP and this therefore a required Feature

snepsnepsnep:


Mack29446 (approved):
Looks good, however I have not tested this

Olrosse (approved):
works great, tested on 0.26 and 0.27, 0.26 it deactivates and 0.27 works like normal

20dka (approved):

PR #296, opened 2023-01-03
Add info about simplified vehicles reducing visual accuracy in tooltip

Desoroxxx:
Extremely tiny and simple PR.

This can be confusing for some people, found the issue when one of my friends noticed he could not see my custom configuration on my car, so precising it in the tooltip will make it clearer.


Mack29446 (approved):
Thank you for this, looks good to me

20dka (approved):

PR #306, opened 2023-01-21
v2/v3 compatibility patch

Starystars67:
This PR adds support for V3 client whilst maintaining compatibility for v2 clients.
I also did a small fix for some settings default and also adjusted the naming convention from nameTagX to nametagX.


Starystars67:
There is however one bug still that I found. That is that when in the main menu the game console is spammed with:
`|E|GELua.|Cannot setCamera, camera not found: freeCamera`
This only happens when using the V3 client. If anyone knows how to fix this one please add to this PR.

Starystars67:
Closing due to age. I will look at how to approach this again in the future.

PR #309, opened 2023-03-08
Improve performance of TCPSocket on MPGameNetwork.

ghost:


ghost:
This adds a queue system which prevents the processing of duplicate data when there are previous matching entries being processed. With this functionality, it also allowed for the use of the `coroutine` library to improve the speed of the `TCPSocket`.

ghost:
Not very satisfied with the pull request now. I believe I should have implemented more features, and also believe that I must finish the API documentation I had been working on before implementing something like this.

Starystars67:
Hi @shawnjb Thank you for raising this PR and sorry for the delay. Most of the core team have work and other matters during the week now.

Please do let us know when you are ready for this to be opened again and we will review then.

lionkor:
It says it improves performance, is there an easy benchmark for this that one could perform?

PR #314, opened 2023-03-22
New Servers Page Features + Fixes

### LINKED ISSUE #284: [Bug] Serverlist playercount filter not accurate
The player count filter "Not full" shows servers which have a current player count over their set limit

Starystars67:
This PR adds a new Featured & Official Servers tab to BeamMP. It also should have a slight performance increase when loading the pages due to it now no longer loading all of the tabs data. It now loads only that current tabs data.

image

~This PR will be dependent on backend version 1.6.3 being made live.~ - Change has been made

The following also still need to be done:
- Write locales for additional languages 
- ~Update sidebar icon~


ghost:
I honestly think rather than creating a new server list you should have a sidebar menu where the user may toggle an option to display featured servers.

Starystars67:
> I honestly think rather than creating a new server list you should have a sidebar menu where the user may toggle an option to display featured servers.

This might be something in the future. I know that there is consideration towards a full rework of the UI which would also take into account redesigning the servers page too.

Mack29446 (dismissed):
please note i have not tested this

PR #321, opened 2023-03-27
Adds a Custom Role API in and for MPVehicleGE's onPreRender function

OfficialLambdax:
BeamMP lacked the feature to script in custom roles for player vehicles. Custom colored roles where reserved for special members. This changes this case. A scripter can now create custom roles and also add custom colors to them. Additionally to that, it is now possible to give specific Vehicles different Names and it is possible to hide Specific vehicle tags.

All custom set Roletags and Names have a * in front of them to indicate that they are custom, which is made to prevent impersonations.

Example. A player under the actual name "beamcruisebot" has spawned in 3 Cars.

One of his cars got a custom orange tag. Visible too is a custom long name tag indicated with a *

Same role but with the short tag option enabled

Here the [*LONG TAG] is removed, but the name of the vehicle is changed. Indicated with a *

Here are three cars given three different roles.
The first has the orange background role with a [TAG].
The second has a purple background role with no [TAG].
The third has a hide playertag role, which removes the entire tag draw for that vehicle

The functions used for this are callable from outside of MPVehicleGE, as they are ment for Scripters use.
MPVehicleGE.createRole() - creates a role
MPVehicleGE.removeRole() - removes a role
MPVehicleGE.setVehicleRole() - gives a vehicle a previously created custom role
MPVehicleGE.removeVehicleRole() - removes a previously given role to that vehicle

The added functions have headers to tell the user how to exactly use them.

POC. Will let you test every function out on the fly via a server side script.
[testscript.zip](https://github.com/BeamMP/BeamMP/files/11081855/testscript.zip)

Custom Roles that are given to Vehicles, overwrite the actual userrole of that Vehicle until its removed again. So a Early Access member for example will have its purple tag unless its overwritten with a Custom role and they will get their EA role back once the custom is removed from the vehicle.

It is not possible to give default roles specified in the roleToInfo table to a vehicle. Meaning that Staff, Developers, EA or youtuber roles are not inheriteable.

One last thing to make clear. With this it is not possible to give entire Players a custom role so that all his vehicles will have the same role on spawn automatically. No the Scripter needs to give each of the players vehicle the role on spawn himself. A simple for loop function is all it takes to do that for the scripter tho.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -41,14 +41,11 @@ local roleToInfo = {
 	['MDEV']	= { backcolor = ColorI(194, 055, 055, 127), tag = " [BeamMP Dev]", shorttag = " [Dev]" }
 }
 
---[[
-	Format - same as roleToInfo
-	Just contains the Custom roles created with createRole()
---]]
+--[[Format - same as roleToInfo
+	Just contains the Custom roles created with createRole() --]]
 local custom_roleToInfo = {}
 
---[[
-	Format
+--[[ Format
 	
 	{"X1-Y1":
 		{"Role":"ROLENAME",
@@ -58,8 +55,8 @@ local custom_roleToInfo = {}
 	}
 	
 	This Format allows for custom Playerroles and Displaynames for each Multiplayer Vehicle.	
-	A defaultrole is "BLANK", which disables the name tag drawing for individual cars
---]]
+	A defaultrole is "BLANK", which disables the name tag drawing for individual cars --]]
+
 local custom_vehicleRoles = {}
 
 local simplified_vehicles = {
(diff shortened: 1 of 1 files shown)


OfficialLambdax:
This allows server admins to give themself a colored Admin tag, their moderators Moderator tags, they could honor their Donators etc..

Or other scripted servers could hide specific Vehicle nametags for certain circumstances. For example CopsNRobbers could hide the Tag of a Robber when a Cop is close and unhide it once the Robber is further away again. Or Its possible to update the Usernames in a race to reflect the Position of the Player. Like "1st. Player1", "2nd. Player2"... Or a RP server could decipher the playertag name until a officer has got the identity of the driver, which adds immersion and gameplay. Imagination can go wild

Starystars67 (approved):
Hey, This is a nice bit of work that you have done here!

Nice commenting too.

Thank you for helping to contribute to the project.

Anonymous-275 (approved):

lionkor:
Id recommend taking a look at the other commit messages, and styling your messages accordingly (e.g. naming them with more "why" information, or squashing them into one if they are one atomic change, or making more atomic ones).

Thats all I got, I cant really speak to the Lua itself

PR #322, opened 2023-03-28
Function headers since the BeamMP Wiki is pretty much empty

OfficialLambdax:
**Draft because not finished yet**

This creates Headers for all Functions that a Scripter can use. It also will add a small format documentation for the players and vehicles table.

All entries that have a %#% mean that i dont know yet what they are actually doing or returning (either at all or just their format).

One question already. The func setOwn(), is this one actually ment to be used by scripters?


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -121,13 +178,15 @@ local function getPlayerByName(name)
 	end
 end
 
-local function localVehiclesExist()
```
OfficialLambdax, line 126: moved to another location, because it cannot be used by a scripter.

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -495,9 +710,6 @@ function Vehicle:onSerialized()
 	return t
 end
 
-local function getPlayers() return players end
-local function getVehicles() return vehicles end
```
OfficialLambdax, line 501: both moved up as they are callable by a scripter

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -11,8 +11,6 @@ local M = {}
 -- ============= VARIABLES =============
 local lastResetTime = {}
 local oneSecCounter = 0
-local vehiclesMap = {}
-local distanceMap = {}
```
OfficialLambdax, line 15: moved down to add format describtions

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -76,11 +56,130 @@ local settingsCache = {
...
+		DisplayName: (String)                              - Contains the Name to display for this vehicle
+		},
+	serverVehicleID_N	
+	}
+	
+	This Format allows for custom Playerroles and Displaynames for each Multiplayer Vehicle.	
+	A defaultrole is "BLANK", which disables the name tag drawing for individual cars --]]
+
 local players = {}
+--[[ Players Format
```
OfficialLambdax: Multiple Function headers point to these Format docs here. This way the Tables contents are just specified once instead of in every function.

OfficialLambdax:
Each function that can be used by a Scripter in a senseful way should now have a Header. Only func missing is setOwn() as i dont think that its a function that is usefull for a Scripter.

Required Variables where also documented, so that everyone can see how the script wide global variables, that are retrieveable, are structured.

Only point i got open there is players.vehicles as of what this table actually contains. Otherwise please review, i hope it got it all

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -41,24 +39,6 @@ local roleToInfo = {
 	['MDEV']	= { backcolor = ColorI(194, 055, 055, 127), tag = " [BeamMP Dev]", shorttag = " [Dev]" }
 }
 
---[[Format - same as roleToInfo
-	Just contains the Custom roles created with createRole() --]]
-local custom_roleToInfo = {}
```
OfficialLambdax, line 48: moved and retyped the format doc

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -76,11 +56,130 @@ local settingsCache = {
...
+	gameVehicleID_N: ..
+	}
+--]]
 
--- ============== MAP HELPERS ==============
+local distanceMap = {}
+--[[ DistanceMap Format
+	
+	{
+	gameVehicleID_1: (Float),                               - Distance from the Players Point of View to this Vehicle in %#%
```
OfficialLambdax: Im not sure what the float represents. My best guess is meters.
It is defined here https://github.com/BeamMP/BeamMP/blob/development/lua/ge/extensions/MPVehicleGE.lua#L1469

20dka: this is correct

lionkor:
For whoever reviews & merges this: Squash to one commit

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -136,12 +277,32 @@ local function isOwn(gameVehicleID)
...
+                   : #
+                   : #
+    Return values .: # IS THIS FUNCTION ACTUALLY MENT TO BE USED BY SCRIPTERS?
+        if success : #
+        if error   : #
+    Remarks .......: #
+    Example .......: No
+-----------------------------------------------------------------------------------------------------------------------------------]]
 -- SET IF A USER OWNS A VEHICLE
 local function setOwn(serverVehicleID, own)
```
OfficialLambdax: Is this function to be used by a scripter?

OfficialLambdax: i guess so then

OfficialLambdax:
please review and merge. I want to work on other parts in this file

lionkor:
My $0.02 would be: Has the [LDoc](https://github.com/lunarmodules/LDoc)/LuaDoc format been considered instead?

Starystars67:
Given this is one file at this point in time I also feel that it would be appropriate to use a recognised documentation system so that it can be used down the road for any docs site too. Please can this be formatted according to [LDoc](https://stevedonovan.github.io/ldoc/manual/doc.md.html) @20dka Thoughts?

OfficialLambdax:
Will convert to ldoc's format

OfficialLambdax:
Example doc will be posted in the next comment

I had to deal with 3 problems.

- **The functions to be used by scripters where all defined as locals**

`local function example() end`
This caused the doc creator to not display the functions unless the -a arg was given. But even then the functions where declared as local function which caused two issues. The first is that a potential consumer of the documentation would get confused because they are declared as locals. The second is that we may wanna document the actual local functions in the future too and when this becomes the case then these functions will be mixed with these that are to be used by a scripter, which will make it confusing.
I have fixed this by removing all local declarations from these functions. This way the doc is presented right and the mod still works.
`function example() end`
Tested the dev build on CarRP and a Vanilla server.

- **The LDoc format doesnt seem to support table in table documentation**

`table["alpha"] = {}`
There are ways to document what a table has
--- Table Describtion
-- @table tableName
-- @tfield string alpha Contains a random String
-- @tfield table beta Contains random Strings
and this example beta table cannot be described in a pretty way, atleast i havent found a way.
So what i ended up doing, was to create documentations for each table that is in a table

v Main Table that has a beta entry that is a table
--- Table Describtion
-- @table tableName
-- @tfield tableName_beta beta Contains random Strings

v Beta table documentation
--- (in table) Beta table Describtion
-- @table tableName_beta
-- @tfield string randomString_1 Contains a random String
-- @tfield string randomString_N ..

this might not be pretty code wise, but it works for the doc creator and it was required for the players and vehicles table as they have a couple more entries.



- **There seems to be a bug within LDoc**

you can enter word-links in specific areas that when you click them moves the user to the apropiate function or table. So if you click the "tableName_beta" on the left of beta from above then it would bring you to the table in question. This is usefull for functions that return complex tables. Like for example getVehicles() returns the whole vehicle table. by using
-- @treturn vehicles
the user can click "vehicles" in the doc and get teleported to the vehicles table documentation. Which is very neat as the user can instantly see the contents of the returned table.



However, this doesnt work when the given word-link contains a underscore (_). When user click a word-link like this then he gets teleported to another but similiar documentation. Which is weird. Either this is a bug or me not knowing how to better declare.

OfficialLambdax:
[doc.zip](https://github.com/BeamMP/BeamMP/files/11133544/doc.zip)

Starystars67 (dismissed):
Great, thank you!

Starystars67:
I see what you are saying with the doc generator. One way we can try to resolve some of this is is we change the functions that are exposed to be like the below example:

Before
```
local function Demo () end

M.Demo = Demo
```

After
```
M.Demo = function() end
```

What are everyones thoughts?

OfficialLambdax:
I think what you mean is already the case?
eg. https://github.com/BeamMP/BeamMP/blob/development/lua/ge/extensions/MPVehicleGE.lua#L1735

Starystars67:
Hey, so it is similar to that but not quite. By doing the approach I showed in my previous comment it should remove the `local` detail that is causing issues for you. Just a thought.

OfficialLambdax:
i removed the local from each function thats to be used by a scripter, so it isnt a issue anymore. I dont know what consequences it would have to switch to `M.Demo = function() end`. I imagine that would require quite some code changes everywhere where the function is called.

I think i saw this in another script where there was a function declared
`M.Demo = function() end`
and it then had to be used like this every in the same script
`if M.Demo() then end`
instead of just
`if Demo() then end`

So changing to this format would require quite some changes within this lua

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -105,44 +245,67 @@ local function getServerVehicleID(gameVehicleID)
...
 end
 
--- SET IF A USER OWNS A VEHICLE
-local function setOwn(serverVehicleID, own)
+--- Sets if the given given Vehicle belongs to this Client. Note from Author of this documentation: Using this function could cause issues.
+-- @tparam string serverVehicleID X-Y. Where X is the PlayerID and Y the Players VehicleID
+-- @tparam bool own
+-- @treturn nil
+-- @usage setOwn("0-0", true)
+function setOwn(serverVehicleID, own)
```
Starystars67: This should still be a local function as this is handled upon receiving the data back from the server.

As such it is for BeamMP Mod only.

OfficialLambdax: alr made it local, edited the header slightly and then moved it a bit down

OfficialLambdax:
functions and local functions are seperated in the doc.
[doc.zip](https://github.com/BeamMP/BeamMP/files/11142813/doc.zip)

Starystars67 (approved):
LGTM :)

OfficialLambdax:
whats up wit this? i spend days on this, overhauling the previous headers to the lua doc ones, looking up the internals to see what is what, learning two doc methods and what not. Im gonna close this soon

Starystars67:
Hi @OfficialLambdax, I am really sorry for the delay on this. I have requested review from the lead of this part of the project so that we can have it merged but they seem to be busy. I have re-bumped this internally for their input.

Sorry again.

PR #331, opened 2023-04-08
Fixes Camera Spawning underground in Italy

OfficialLambdax:
Italy does not have the PlayerDropPoints folder in its contents, which makes it impossible for beammp to spawn the camera at the default location. This patch takes the spawnpoints from another file if the first, thats used by the other vanilla maps and modded maps, is not available.


Starystars67 (approved):

OfficialLambdax:
Closed duo to Deer working on a much better fix

PR #334, opened 2023-04-14
Feature IMGUI

Starystars67:
This PR adds in the new IMGUI chat system which was originally created by @vulcan-dev.

Since then I have made some small adjustments and brought it into the newer codebase.

Screenshot for your enjoyment:

image

PR #335, opened 2023-04-14
Feature IMGUI Chat + Server list work

Starystars67:
~This PR replaces https://github.com/BeamMP/BeamMP/pull/334 due to further work.~

~This PR adds in the new IMGUI chat system which was originally created by @vulcan-dev.~
This PR makes some adjustments to the work done by @vulcan-dev.

Since then I have made some small adjustments and brought it into the newer codebase.

Screenshot for your enjoyment:

image


Starystars67:
The new Player ID is also ordered by number.

Olrosse:
found some issues, the emojis like those used on carp just show up as question marks, and pressing the up arrow doesn't show the last message like the UI chat does, that can be quite useful for chat commands

daniel-w0:
> found some issues, the emojis like those used on carp just show up as question marks, and pressing the up arrow doesn't show the last message like the UI chat does, that can be quite useful for chat commands

I thought I added the history, I guess I didn't get around to it. It's really easy to add though.
With the emojis, I don't see any possible solution besides this (https://github.com/ocornut/imgui/blob/master/docs/FONTS.md#using-colorful-glyphsemojis). I don't see them being important anyway. Maybe we could have custom emojis, for example ":carp:" would map to a carp emoji which will would be its own separate image in some folder.

OfficialLambdax:
For the issues with the emojis. Maybe leave the current chat ui in the mod and rename it to "legacy chat" or smt like that. That way a client side script could disable the new chat and use the old if they have for example emojis. Or even better, the player can choose which chat they want to use. I personally for example like the current chat

daniel-w0:
> For the issues with the emojis. Maybe leave the current chat ui in the mod and rename it to "legacy chat" or smt like that. That way a client side script could disable the new chat and use the old if they have for example emojis. Or even better, the player can choose which chat they want to use. I personally for example like the current chat

I'm already doing that. There's an option in the multiplater category to enable the new menu.

Issue #285, opened 2022-12-19
[Bug] Rapid accelleration causes emergency brakes and destroyed cars


OfficialLambdax:
First of all this is a game thing and something that the mod just doesnt address, because of it this could be a feature request, but since it causes issues i think its better to flag it as a Bug.

**Describe the bug**
Sorry for poor english, i have troubles describing this bug.

If a player joins a server while there are players on it that are driving fast cars (any car). After the joining player loaded the already present vehicles the mod applies the other vehicles positions, velocities, rotations etc. The problem that arises here comes from the fact that cars spawn without movement. So once the velocity, rotation etc. is applied to the car, and if the velocity or rotation in any given direction is high enough, then that rapidly accellerates the vehicle. The accelleration can easiely be 20g and higher.

This whole issue comes in my perspective down to very High G that get created through the rapid accelleration.
Let me demonstrate what this does to a vehicle. The video shows a scintilla that is reset and then rapidly accellerated. As you can see in the g meter, the forces (if i read correctly) are at 29g. You can visually see that the vehicle gets demolished. It took several car resets to bring the vehicle to a non damaged state as can be seen further into the video.
https://www.youtube.com/watch?v=IfVrWmtgqGE

This exact behaviour happens to every car that is moving fast after the players client was hung (during vehicle spawn/edit/game pause/lag etc.). If the car has a emergency stopping system (ETK-K, Bastion, Vivace, Scintilla...) then the game also often enables that on the affected vehicles and does not stop until the affected vehicle got reset or stopped. This is why we can see vehicles with fully locked brakes and wheels that are smoking over the asphalt.

**To Reproduce**
1. Join a server where there are players that are driving fast cars. Problem is most prominent on low com vehicles like the scintilla.
2. (Alternative) Message me. I can reproduce the bug with the Bot. All it needs is a simple command in the chat and it replicates it instantly as often as you want.

OfficialLambdax:
A better video
https://www.youtube.com/watch?v=YNcbDDI5EBg

In the 1st. Try you can see visual damages to the car
In the 2nd. you can see that the rapid acc worked mostly nicely
In the 3rd. you can see that the vehicles enabled emergency brakes
In the 4h. you can also see emergeny brakes and that the radiator got damaged (the car starts smoking)
In the 5th. ive reseted the car multiple times by hand, but to no avail, because a reset resets the vehicles velocity and rotation to zero and since its rapidly accellerated again, the same issue breaks it again

[...]

PR #339, opened 2023-04-20
Vehicle demolition fix that is caused by game lag

OfficialLambdax:
Needs some polish, further testing and the electrics update done. But overall this seems to work pretty well already. However it does not cover each case. Randomly i still see demolished vehicles

Also the debug print() needs to be removed before merge.


OfficialLambdax:
This fix addresses https://github.com/BeamMP/BeamMP/issues/285

lua/ge/extensions/positionGE.lua
```diff
@@ -9,6 +9,11 @@ local M = {}
 
 local actualSimSpeed = 1
 
+local highVelocityFixPause = {} -- serverVehicleID:os.time()
```
OfficialLambdax, line 12: This is a table that is used to stores serverVehicleID's in that are currently paused from getting the fix applied to them. Yes this table can contain id's from players that are no longer on the server and you may call it dirty. But that is not a issue, because the id is reclaimed again once a new player gets it. So no this is not a memory leak

Starystars67:
Hey @OfficialLambdax , Any update on this?

OfficialLambdax:
> Any update on this?

No Update on this

Starystars67:
Are you planning to change this from a draft? 
I would like to consider getting this merged :)

OfficialLambdax:
> I would like to consider getting this merged :)

I dont. I dislike the solution.. it resets the vehicles when it detects that a rapid vehicle demolition would take place.. but id rather not to. This could be solved better with damage sync. For me this will stay a draft.. or at best become a experimental option

Starystars67:
> > I would like to consider getting this merged :)
> 
> I dont. I dislike the solution.. it resets the vehicles when it detects that a rapid vehicle demolition would take place.. but id rather not to. This could be solved better with damage sync. For me this will stay a draft.. or at best become a experimental option

Understood thank you. Will wait for new launcher to be in use first before revisiting.

Desoroxxx:
> I dont. I dislike the solution.. it resets the vehicles when it detects that a rapid vehicle demolition would take place.. but id rather not to. This could be solved better with damage sync. For me this will stay a draft.. or at best become a experimental option

Damage sync sure does sound interesting but would require a lot of work I asume

Starystars67:
New launcher is now in EA testing.
We can start to come back to the Damage/Node Sync now @OfficialLambdax .

Starystars67:
👀 New activity!

OfficialLambdax:
Im noping out of this one, no matter how often i try, this solution here is at the wrong place. It needs to work in conjunction with ve lua as ve lua determines a trajectory and likely position of the remote vehicle.

PR #341, opened 2023-04-23
fix several hardcoded strings, and new simplified chinese translations

Please-Pick-a-Name:
the clear recent string is also added in the en translation file
this pr includes oiler's pr (hi, idk why im everywhere but whatever)

PR #343, opened 2023-04-28
Create de_DE.json

O1LER:
Created german localization file including newly added strings per #341


Starystars67 (approved):
Thank you for adding and updating this.

PR #345, opened 2023-05-02
Fixed #308

### LINKED ISSUE #308: [Bug] Certain serverlist filters can cause the list to stay empty
**Describe the bug**
When selecting both "Empty only" and "Not empty" serverlist filters at the same time, it will filter out all of the servers, causing the list to be empty

**To Reproduce**
- Go to the serverlist
- Wait for servers to appear in the list
- Activate "Empty only" filter
- Notice certain server to disappear
- Activate "Not empty" filter
- Notice all the servers to disappear from the list

**Expected behavior**
It should not be possible to activate filter combinations that exclude all servers.
If one of the two options is active, the other should be greyed-out.

**Screenshots**
image
image
image

ArmenG888:
Added a if statement to check if CheckIsEmpty and checkIsNotEmpty are true and to change to false


Starystars67 (dismissed):
Thank you for raising this in a PR, I did do something similar where they cancel each other out in another PR: 
https://github.com/BeamMP/BeamMP/pull/335/files#diff-cb8a69ebbaa7c4a0117167c0080f2ce31e561db88232112b1e7fd63d3a293546R352

Lets see what the rest of the team think on this though.

PR #353, opened 2023-06-07
Issue #349

### LINKED ISSUE #349: Move 'Quit' button in session UI app
Inadvertently quitting a server is very frustrating. Unfortunately, the placement of the 'Quit' button in the in-game session UI app - especially when manually queueing events - can easily lead to misclicks to exactly this effect.
This is because the 'Quit' button is right next to the 'Events queued' button, so a slight misclick instantly quits the server without additional confirmation. Even worse is the situation where the queued event(s) disappear the second you click the 'Events queued' button, which removes the button and moves the 'Quit' button to the left, potentially clicking quit instead.

I propose to move the 'Quit' button to the very left of the UI element. This way it's far away from any other button in the session UI and also right next to the server name, which seems logical.
The required code change is trivial.

**Describe alternatives you've considered**
Asking for an additional confirmation before quitting a server would also eliminate this problem. However, it may feel cumbersome for some if the need for a confirmation can not be toggled off in the settings, which in turn makes this solution more complex to implement than the proposed fix.

**Additional context**
Visualization of the proposed change:
move_quit_button_illustration

ArmenG888:
Moved the quit button to the left

### DIFF
--- a/ui/modules/apps/BeamMP-Session/app.html
+++ b/ui/modules/apps/BeamMP-Session/app.html
@@ -70,6 +70,5 @@
 		<div class="block others-block" id="queue-block" style="display: none; padding: 0px" >
 			<button class="buttons" style="background-color: transparent" ng-click="applyQueue()">Events queued: <span id="Session-Queue" title="">0</span></button>
 		</div>
-		
 	</div>
 </div>
--- a/ui/modules/apps/BeamMP-Session/app.html
+++ b/ui/modules/apps/BeamMP-Session/app.html
@@ -55,6 +55,9 @@
 	}
 	</style>
 	<div class="outerDiv">
+		<div>
+			<button type="button" id="quit-button" name="button" class="buttons" ng-click="mpquit()">Quit</button>
+		</div>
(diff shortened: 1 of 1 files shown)


ui/modules/apps/BeamMP-Session/app.html
```diff
@@ -67,8 +70,6 @@
 		<div class="block others-block" id="queue-block" style="display: none; padding: 0px" >
 			<button class="buttons" style="background-color: transparent" ng-click="applyQueue()">Events queued: <span id="Session-Queue" title="">0</span></button>
 		</div>
-		<div>
-			<button type="button" id="quit-button" name="button" class="buttons" ng-click="mpquit()">Quit</button>
-		</div>
+
```
Starystars67: Unnecessary new line. Please remove :)

Starystars67 (changes requested):
Please remove the newline and then I will approve for you :)
Thank you for contributing!

Starystars67 (approved):
Great thank you :)

PR #369, opened 2023-07-18
Moved blob drawing out of hideNameTags

### LINKED ISSUE #303: [Bug] Placeholder orbs are hidden with nametag hiding option
Yes

**Describe the bug**
Hiding nametags also hides placeholder orbs

**To Reproduce**
1. Connect to a random server
2. Dont load events
3. Hide player nametags
4. Observe that also placeholder orbs vanish

**Expected behavior**
Placeholder orbs should be always visible, since they not only show a players position but also that there is an event to be loaded (or other things depending in orb color). Nametags and orbs should either have their own visibility setting, or orbs should be always visible independent from nametags.

**Screenshots**
The video was made to show an issue with nametag fade out, but at the beginning you can see that the hide nametag option also hides the orb, at the time i wasnt aware that this itself might also be an issue

https://youtu.be/zSVY13sxxBA

O1LER:
This fixes #303

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1465,6 +1465,30 @@ local function onPreRender(dt)
 			if not v.position then goto skip_vehicle end -- return if no position has been received yet
 			local pos = Point3F(v.position.x, v.position.y, v.position.z)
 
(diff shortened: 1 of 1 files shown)


Starystars67 (approved):
Looks good to me :)

PR #370, opened 2023-07-21
Added simplified traffic identifiers for all cars

O1LER:
- Lansdale was missing since 0.29 update
- Rest of the cars dont have simplified versions by beamNG yet, however, i plan on creating some more for CaRP and will be releasing them as a mod to the repo, so anyone interested could add them to their server. Or, depending on filesize, we could add them directly to beamMP

I tested this on multiple public servers, official and modded ones. If the simplified version of a car is not available, the game will use the the body that is defined in the cars main jbeam as being the standard part. Most of the time this will be the most basic body, so it is already going into the right direction


Starystars67 (approved):
Thank you @O1LER <3

PR #371, opened 2023-07-22
Fix nametag dont fully hide option

### LINKED ISSUE #244: [Bug] don't fully hide nametags broken
"don't fully hide nametags broken" option is broken, nametags do fully disappear. its possible this is a settings related issue, as 0.26 had some of those

O1LER:
Fixes #244


Starystars67 (approved):
Looks good but not tested

OfficialLambdax (approved):
This is the correct setting name according to
https://github.com/BeamMP/BeamMP/blob/development/ui/modules/options/multiplayer.partial.html#L221

And it works

Starystars67 (approved):

PR #374, opened 2023-07-24
Launcher v2.1 compat

Starystars67:


Starystars67:
This has all been confirmed working and tested with both launcher v2.0.X and v2.1.
Players I tested with reported no issues in playing with me.

OfficialLambdax (changes requested):
Please provide the v2.1 launcher. Cannot test without

OfficialLambdax:
Wrong milestone lmao. Why is this 4.9? 4.8 is not Stable yet, isnt this just a 4.8 combat with launcher 2.1?

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -42,8 +42,24 @@ C  -> The client asks for the server's mods
 
 
 -- ============= LAUNCHER RELATED =============
+
+-- Sends data through a TCP socket or IPC to the launcher depending on if the launcher is V2 or V3 Networking.
+-- If V3 Networking is available, it will be used, otherwise V2 Networking will be used.
+-- @param s string containing the data to send to the launcher
 local function send(s)
+	-- First check if we are V3 Networking or not
```
OfficialLambdax: V2.1 Networking

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -346,26 +373,60 @@ local updateUiTimer = 0
...
+
 local function onUpdate(dt)
 	pingTimer = pingTimer + dt
 	reconnectTimer = reconnectTimer + dt
-	updateUiTimer = updateUiTimer + dt
-	heartbeatTimer = heartbeatTimer + dt
+	if status == "LoadingResources" then
+		updateUiTimer = updateUiTimer + dt
+	end
+	if not MP then -- This is not required in V3
```
OfficialLambdax: V2.1.. and more. not going to point to all

Starystars67:
> Please provide the v2.1 launcher. Cannot test without

Launcher is available in the Contributor chat and was shared with the staff team previously too.

> Wrong milestone lmao. Why is this 4.9? 4.8 is not Stable yet, isnt this just a 4.8 combat with launcher 2.1?

This is 4.9 because it is not a patch version change. This is a series of changed that is quite substantive. It would have been a major version change but then I modified it further to be backwards compatible.

lua/ge/extensions/MPCoreNetwork.lua
Starystars67: This has now been fixed, thank you.
This happened because it was going to end up being a V3 but then we held back some other work for the time being.

OfficialLambdax (approved):
This works well on my side, no errors no runtime issues and i also dont see any code errors with either v2.0 or v.2.1. Someone may want to make long runtime tests, also on servers that are heaviely scripted like CaRP.

Only issues i see is that the server list is requested three times and that switching between servers while in one, is causing issues with mod loading. But those two have nothing todo with this compat.

PR #373, opened 2023-07-23
Add keybinds for applying queued events and Toggling nametags.

2PXL:
adds function to toggle nametag with a custom bind (part 2)

this also adds a bind to queue events

PR #378, opened 2023-07-27
toggleNicknames() Fix

OfficialLambdax:
The function toggleNicknames introduced in https://github.com/BeamMP/BeamMP/pull/373
overwrites script side set hidden Nicknames. So if a Client side script force disables nicknames by calling hideNicknames() then this toggle that is accessible via a keybind, will reenable it.

This fixes that.

Draft because: Untested code

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -25,7 +25,7 @@ local vehiclesToSync = {}
 local sentPastVehiclesYet = true
 local queueApplyTimer = 0
 local isAtSyncSpeed = true
-local hideNicknames = false
+local hideNicknamesToggle = false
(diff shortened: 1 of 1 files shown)


Starystars67 (dismissed):
No issues with the code looking at it.
Definitely a better way to do it, Thank you.

Starystars67:
@OfficialLambdax are you ready for this to be merged?
If so please can you change it from a draft?

OfficialLambdax:
> @OfficialLambdax are you ready for this to be merged? If so please can you change it from a draft?
Now you can

@2PXL you introduced a new variable called hideNicknames. This name was already taken by a function. Setting this variable would make the function inaccesible

Starystars67 (approved):

PR #379, opened 2023-08-02
Fixes the Fadeout Close Vehicles Option

OfficialLambdax:
if that option is tickes then it will cause lots of error logs when the to be faded vehicle doesnt exists yet, because of a missing check.

Additionally the fade out distance was hooked to the same distance that nametags are faded out. Since players can have this setting into the hundreds of meters.. this can cause any vehicle in the players vicinity to not be or barely visible. The fade distance is now hardcoded (until a setting is provided) to 20 Meters.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1747,9 +1747,9 @@ local function onPreRender(dt)
 					dist = string.format(" %s %s", tostring(mapEntry), unit)
 				end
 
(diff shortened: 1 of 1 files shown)


Starystars67 (approved):
Hey, sorry for the delay. I was away for a few days.

Looks good to me but I have not tested in game. Have you tested it?

OfficialLambdax:
> Looks good to me but I have not tested in game. Have you tested it?

Yeah i tested it

PR #396, opened 2023-10-24
Fixes #388 -> MPGameNetwork attempts to index a function

### LINKED ISSUE #388: [Bug] MPGameNetwork attempts to index a function
Yes

**Describe the bug**
image

**To Reproduce**
1. Attempt to join a local BeamMP server

**Expected behavior**
I should load into the server instead of being stuck in the loading screen

OfficialLambdax:
Untested. The lua mod introduces "nop" as being the default value for the var TCPLauncherSocket. I have no idea what this is. But this var is either "nop" or a cfunction from the socket lib. And it looks like that the expression _not nop_, is false, which appears to cause this bug https://github.com/BeamMP/BeamMP/issues/388

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -58,7 +58,7 @@ local function send(s)
 	
 	end
 	-- Else we now will use the V2 Networking
-	if not TCPLauncherSocket then return end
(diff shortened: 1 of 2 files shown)


Starystars67 (approved):
Interestingly this does not seem to happen otherwise. 

Testing this in-game there are no errors reported and all seems to work as expected.

PR #398, opened 2023-11-01
Fix chat resetting due to UI app not being loaded still sometimes

Starystars67:
Does what it says on the tin!


OfficialLambdax (approved):
It just works

PR #399, opened 2023-11-02
fix playerlist and chat settings menus not displaying in new chat

StanleyDudek:
unique id needed in imgui buttons or else all buttons share same id

image

image


OfficialLambdax (approved):
It Just Works

Starystars67 (approved):
The missing return M is also likely to not help haha, not sure where that one went!

Thanks for looking into this though <3

PR #406, opened 2023-11-08
Fixes broken downshifts in automatic like gearboxes

OfficialLambdax:


Starystars67 (approved):
Visually all looks good but not tested in-game.

Can you confirm if it has been tested?

OfficialLambdax:
Yeah i tested it with all gearbox

PR #410, opened 2023-12-06
keep text input cursor inside text input field after sending message

StanleyDudek:
very small change, but this is QoL for sending multiple chat messages without having to use the mouse to refocus

### DIFF
--- a/lua/ge/extensions/UI.lua
+++ b/lua/ge/extensions/UI.lua
@@ -61,7 +61,8 @@ M.defaultSettings = {
         inactiveFade = true,
         fadeTime = 2.5,
         fadeWhenCollapsed = false,
-        showOnMessage = true
+        showOnMessage = true,
+        keepActive = true
     }
 }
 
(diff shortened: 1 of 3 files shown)


StanleyDudek:
Sorry, I forgot to mention that this is for NEW imgui chat

OfficialLambdax (dismissed):
Does the job as described <3

Starystars67:
Would it be possible to have this as a configurable setting?
This way a user can choose if they want it to remain focused or not.

StanleyDudek:
> Would it be possible to have this as a configurable setting? This way a user can choose if they want it to remain focused or not.

Yes, this now has an option for keeping the cursor active in the chat, works as you requested, looks like this:

image

NeotericGamer98:
Looks very impressive so far.

Starystars67 (approved):
Nice work @StanleyDudek ,

Thank you for your contributions.

PR #411, opened 2023-12-06
Create release.yml for Automated Release Artifact Creation

Starystars67:
This Action adds automated release artifact creation and addition

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,33 @@
+name: Create Release Zip and SHA-256
+on:
+  release:
+    types:
+      - created
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out the code
+      uses: actions/checkout@v2
+      with:
+        ref: 'development'
+    - name: Zip specific files and folders
+      run: |
+        zip -r beammp.zip icons lua mp_locales scripts settings ui vehicles CONTRIBUTING.md CODE_OF_CONDUCT.md LICENSE.md README.md
+      working-directory: ${{ github.workspace }}
+    - name: Generate SHA-256 hash
+      id: sha256
+      run: |
+        sha256sum BeamMP.zip | awk '{print $1}' > beammp.zip.sha256
+      working-directory: ${{ github.workspace }}
+    - name: Upload artifacts
+      run: |
+        curl -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Content-Type: application/zip" \
+          --data-binary "@beammp.zip" \
+          --url "https://uploads.github.com/repos/${{ github.repository }}/releases/${{ github.event.release.id }}/assets?name=beammp.zip"
+        curl -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Content-Type: text/plain" \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
+name: Create Release Zip and SHA-256
+on:
+  release:
+    types:
+      - created
```
lionkor, line 5: I would instead suggest

```yaml
on:
  push:
    tags:
    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
```

this means you push a vX.Y.Z tag it will run... see the next comment for the next step

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
```
lionkor, line 6: you could add a create-release job, so it would create a release on tag creation
``` yaml
jobs:
  create-release:
    runs-on: ubuntu-latest
    name: Create Release
    outputs: 
      upload_url: ${{ steps.create_release.outputs.upload_url }} 
    steps:
      - name: Create Release
        id: create_release
        uses: actions/create-release@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          tag_name: ${{ github.ref }}
          release_name: ${{ github.ref }}
          draft: false
          prerelease: true
          body: |
              Here goes a description of this release, bla bla
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Upload artifacts
+      run: |
+        curl -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Content-Type: application/zip" \
+          --data-binary "@beammp.zip" \
+          --url "https://uploads.github.com/repos/${{ github.repository }}/releases/${{ github.event.release.id }}/assets?name=beammp.zip"
+        curl -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Content-Type: text/plain" \
+          --data-binary "@beammp.zip.sha256" \
+          --url "https://uploads.github.com/repos/${{ github.repository }}/releases/${{ github.event.release.id }}/assets?name=beammp.zip.sha256"
```
lionkor, line 32: Finally here, instead of curl, you can do

```yaml
      - name: Upload Release Asset
        id: upload-release-asset 
        uses: actions/upload-release-asset@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ needs.create-release.outputs.upload_url }}
          asset_path: ./BeamMP.zip
          asset_name: BeamMP.zip
          asset_content_type: application/zip
```

and the same thing for the sha256.

lionkor (changes requested):
I personally prefer the approach of creating a tag, and then pushing that tag, which subsequently creates a release with everything needed. That way anyone can make a release with push permissions to the repo, but all releases look uniform and so on.

You could limit it to just tags created on develop, but as seen in the server recently, sometimes you want to deploy a hotfix release from an unmerged branch.

You have to decide which workflow you like better, but either way maybe the step about using upload-release-asset or whatever is easiert to use than curl (?)

lionkor (approved):

PR #448, opened 2023-12-16
[New experimental] Positional update inconsistency smoother

### LINKED ISSUE #449: [Bug] Buffout packet receive inconsistencies
timeline

Unstable connections lead to inconsistent positional packet executions. Which lead to visible vehicle warping. Vehicles warp back, forth and side ways. While i dont know how to fix the third case, i think we can buffout the second, by offsetting when to execute a pos update when it is received early or late.

Demonstrative Example from Leviathan of the BeamTC community
https://github.com/BeamMP/BeamMP/assets/91626975/51ab4ee2-9e8e-4991-891f-bcc0768950cf

OfficialLambdax:
see issue https://github.com/BeamMP/BeamMP/issues/449
- only runs if experimental option is ticked


OfficialLambdax:
Here are some early results of the, by now revised, pos exec offsetter (just going to call it smoother).
The `.csv` is written using a temporary new function within positionGE.lua which is called each time `applyPos()` is called.
- Each row represents a PlayerID
- Each Number the time it took since the last pos update execution from `applyPos()` in ms
- Data comes from players with ping ranging from 100 to 400 (Official Derby Server)

The Smoother can be enabled/disabled via the experimental options.
current_status

The smoother tracks and ticks each vehicle independently based on their median interval time. We can receive data every 32ms (optimal case) or in some cases every 50ms. The smoother tries to find the median from the recent history of intervals, and will tick accordingly.
- eg `30.7566, 39.4574, 60.5794, 30.3344, 43.838, 32.2568, 51.4615, 62.6635, 31.6746, 48.3338` means a median of 43.1356. Packets that arrive after the median tick, will be executed directly.

Note to myself. i will have to see if ticking by the highest 10% intervals is better.

OfficialLambdax:
Using a different median timer that tracks the direct `packet <-> packet` intervals instead of `last_exec <-> packet` intervals while also allowing lower then 32ms intervals into the calcula, appears to be showing better results. Had 100 to 300 ping players on the server.
current_status

Starystars67:
This is great and I cant wait to see it in practice!

lionkor:
I'd like to ask for 2 people to review and test this instead of the usual one

Starystars67:
> I'd like to ask for 2 people to review and test this instead of the usual one

Agreed and once @OfficialLambdax is ready for it to be I will be one of them :)

OfficialLambdax:
It is not exactly easy to get this tested. You have to find a player that has a inconsistent ping (eg. fluctuating between 60 <-> 120) and where their vehicle is warping. Then you can switch on the experimental option `Enable Positional execution smoother` and see the effects it has. This pr cannot solve desync comming from high ping players (>300) and neither can it fix the situation where there is a loss of packets.

This is vehicle warping. Can exist in extreme instances like in this video, or in small instances where the vehicle is looking like its braking, accelerating, braking, accelerating while it actually just drives at the same speed
*Provided by a player named Leviathan from the BeamTC community*
https://github.com/BeamMP/BeamMP/assets/91626975/3754c9ab-717c-4b53-9391-adbc90ecee7b

Duo to the nature of how hard it is to reproduce vehicle warping from the getgo, this would be a experimental option.

Starystars67:
I ran this PR today and experienced no issues in the code / no errors.

It was difficult to determine exactly if this was having much of an affect on the sessions I joined.
That said as outlined above this is something that is only going to have any real affect on players with unstable pings.

Desoroxxx:
I often play with my friends who have unstable ping, and this issue occurs a lot for us, which is the main reason we always pay attention to never be behind each other's, I would be very interested in testing this.

Would this also improve situations where when a player is doing the events and thus have their game freezed we still see them moving?

O1LER:
> Would this also improve situations where when a player is doing the events and thus have their game freezed we still see them moving?

No, since loading an event essentially stops the vehicle simulation. If theres no movement of a players car, no matter the reason, how could it be still moving for others. Or it should not keep on moving, since that would create a difference between the cars position for the car owner and everyone else

O1LER:
Been testing this for about an hour. I guess its working, hard to really tell. There might be some rubberbanding now, but that is to be expected with any kind of smoothing

OfficialLambdax:
@Desoroxxx 
> Would this also improve situations where when a player is doing the events and thus have their game freezed we still see them moving?

No, for that your game would need to know when another player is freezing, this is not covered by this pr

> I would be very interested in testing this.

Could you hit me up on discord? (tag is neverless.) im looking to get this tested through races that have players with unstable connections, just couldnt get around it yet

Starystars67 (approved):

PR #450, opened 2023-12-18
Fix persistent big map icons

### LINKED ISSUE #430: [0.31] Big Map issue
image

snepsnepsnep:
Partially fixes https://github.com/BeamMP/BeamMP/issues/430
The icons will no longer persist but the map still won't open properly with no vehicles spawned, or if the camera is on a nil vehicle.

Removes some unused/unreferenced code.


Starystars67 (approved):
Indeed after testing multiple times I am now unable to recreate the icons issue.

Nice work!

PR #456, opened 2023-12-20
Set "Autosave Unicycle" as the new default functionality

### LINKED ISSUE #455: [Bug] Beamlings Customisation does not save
**The Issue**

When selecting a customisation for the beamlings and then get into a vehicle, the character resets to default in a server with multiple unicycles.

**To Reproduce**
- Load into a server with another person and customise your character. 
- get into a vehicle and get out again.

**Expected behavior**
The characters are supposed to keep their customisation when getting in and out of vehicles.

**Additional context**
I was not there when they issues were found, these were reported by the beaming YouTube recording team.

OfficialLambdax:
The "manual over the menu" as default functionality is confusing for players. Players expect that the default unicycle is always the last used. Since this feature is already present in beammp, but just not the default, swapping a boolean makes it the default functionality.


OfficialLambdax:
Players who already ran the prerelease will already have the `unicycleAutoSave` option set to false. It needs to be announced in the full release that those have to toggle the autosave manually on in the settings.

Starystars67:
Hi @OfficialLambdax thanks for raising this fix, Indeed it seems and important one.
We can put an announcement out to the EA members who have this that they should do that setting change.

Starystars67 (approved):

PR #474, opened 2024-01-01
Fix unsupported powertrain device handling

### LINKED ISSUE #475: [Bug] Unsupported powertrain devices are not handled
`applyGear()` in `MPInputsVE.lua` does not handle unsupported powertrainDevices. This leads to the game deactivating a vehicle with a unsupported device.

OfficialLambdax:
Some mods introduce their own custom powertrain. Some are incompatible to the way beammp shifts a gearbox. Unsupported devices where not handled so far which lead to the vehicle to be deactivated by the game. This pr wont introduce missing powertrains, but rather just prevents the execution of `applyGear()` in `MPInputsVE.lua`

### DIFF
--- a/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
+++ b/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
@@ -17,6 +17,8 @@ local lastInputs = {
 	c = 0,
 }
 local remoteGear
+local unsupportedPowertrainDevice = false
+local unsupportedPowertrainGearbox = false
 -- ============= VARIABLES =============
 
 local translationTable = {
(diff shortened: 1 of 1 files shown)


Starystars67:
Do you have any examples of these mods so we can test this?

OfficialLambdax:
Feel free to join CaveRP servers. This issue can be seen with airplane mods (albeit probably not all). However another player needs to fly it for this bug to appear on your client

StanleyDudek (approved):
This works as described, fixes large roller, large spinner, and various aircraft

Olrosse (approved):
from testing it seems to work as expected

PR #479, opened 2024-01-02
Improve the kick dialog (onAuth)

NenoxAG:
This PR fixes console spam that occurs when you get kicked and also adds support to multiline kick messages using "^p". This character was chosen, because it is also used in the description field of the server list.

### DIFF
--- a/ui/modules/multiplayer/multiplayer.js
+++ b/ui/modules/multiplayer/multiplayer.js
@@ -275,10 +275,17 @@ function($scope, $state, $timeout, $mdDialog) {
 
 	$scope.$on('LoadingInfo', function (event, data) {
 		if (document.getElementById('LoadingStatus').innerText != data.message) console.log(data.message)
+
+		// Escape all html
+		data.message = data.message.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+		
+		// Replace ^p with <br>
+		data.message = data.message.replace(/\^p/g, "<br>");
+
 		if (data.message == "done") {
-			document.getElementById('LoadingStatus').innerText = "Done";
+			document.getElementById('LoadingStatus').innerHTML = "Done";
 		} else {
-			document.getElementById('LoadingStatus').innerText = data.message;
+			document.getElementById('LoadingStatus').innerHTML = data.message;
 		}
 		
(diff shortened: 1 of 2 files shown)


ui/modules/multiplayer/multiplayer.js
```diff
@@ -275,10 +275,17 @@ function($scope, $state, $timeout, $mdDialog) {
 
 	$scope.$on('LoadingInfo', function (event, data) {
 		if (document.getElementById('LoadingStatus').innerText != data.message) console.log(data.message)
+
+		// Escape all html
+		data.message = data.message.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+		
+		// Replace ^p with <br>
+		data.message = data.message.replace(/\^p/g, "<br>");
```
lionkor, line 283: Hi, this is not actually as internally consistent as you may think. So, `^p` is used for newlines, sure, but that's for *user-editable* fields like the description.
Kick messages are user-editable, but via scripting or console only. The console doesn't support newlines, and if you manage to put one in, it'll send it as `\n` (0x0a), not `^p`. Scripts could reasonably send one with a newline, but again, not `^p`.
So, supporting this would mean server changes in sync with this feature, which IMO isn't worth it.

Would it work to use normal newlines `\n` (0x0a) here?

lionkor (changes requested):

ui/modules/multiplayer/multiplayer.js
NenoxAG: The problem with using `\n` (0x0a) is that the `:receive()` function will only read one line terminated by an LF (0x0a) character, as per the documentation: https://lunarmodules.github.io/luasocket/tcp.html#receive 
When using a `\n` in the kick message, it throws an error because the line gets cut off, and the next receive operation encounters a possibly not existing character as the code value. This is the reason why the console spam (e27ffd9) exists.

I think it would be really helpful if it were possible to send multi-line onAuth kick messages, a use-case for example would be, displaying ban information in a structured way.

lionkor: Okay, opening an issue for that since the newline is not a good terminator.

lionkor:
This won't be merged right now because it has no support in the server, and is only a bodge. The real fix is to allow `\n`, but if someone makes a server PR to generate `^p` on newline, we can merge this.

PR #482, opened 2024-01-03
[Suggestion] Snowman texture removal

OfficialLambdax:
The main game features their own snowman textures. Removing ours from the mod doesnt break the snowman model
uikluioöliolö9op

Since the textures are 1mb in total size, the removal shrinks the mod size considerable (from 1.4mb to 1mb). I do not know why the textures where included in the mod in the beginning 3 years ago, but the focus is now on the beamlings and i dont see why these textures need to be kept if the game itself also contains them (they are different tho)


Please-Pick-a-Name:
if the textures in beammp are the same as the ingame ones 3 years ago, then ig its just whoever did the unicycle werent confident about vehicle modding enough
if they were never the same, then ig we can have a comparison here and see what exactly are the differences?

Starystars67:
@VanilleVaschnille can you confirm if you need them as part of the Beamlings?
Thanks

VanilleVaschnille:
> @VanilleVaschnille can you confirm if you need them as part of the Beamlings? Thanks

I would like to keep the UwU snowman as it is a staple of the community. Removing it would cause unneeded unrest on the community. People would be quite upset.

PR #486, opened 2024-01-05
[Update] Positional execution smoother

### LINKED ISSUE #449: [Bug] Buffout packet receive inconsistencies

OfficialLambdax:
Adds refinements for the smoother


Starystars67:
Hi @OfficialLambdax, any further updates on this one?

OfficialLambdax:
Buffing out receive inconsistencies still shows being a valid approach to combat one cause of rubberbanding, that being the kind that comes from executing position packets to quickly or in the wrong order.

Tho im out of ideas on how to move this further and just want to get this PR out of the way. Feedback had been varied, some said that this helps them alot with door to door close racing, others said it made it worse when everyone has a very low ping. I personally havent seen a rubberbanding case in a long while (but im also inactive).

I hope this pr's progress can make the experimental a bit better.

Olrosse:
i tried enabling this when we where racing around the oval at motorsports playground a while ago when the rubber banding was really bad, it was a high ping US West server with 10 ish people, i didn't really spot an obvious improvement if i can remember correctly so i disabled it again, so i think we would need to do some more controlled A/B testing,

I'm not sure if this is the best solution, timestamps on all packets with a buffer like Stefan has suggested is probably the way to go, but have you tried doing this in PositionVE rather than GE? might be a bit better on performance and it could get rid if another frame of inconsistency

OfficialLambdax:
I dont know how the rubberbanding in your case was caused, this here targets that cause #449
Your driving close to someone and suddenly they warp forward or backwards or even teleport like seen in this video
https://github.com/BeamMP/BeamMP/assets/91626975/51ab4ee2-9e8e-4991-891f-bcc0768950cf

And no i didnt try to move this to VE lua. All i know is that the interpolation done in VE cant handle this case of rubberbanding.

Olrosse:
i found the footage from that session, and looking at it it worked a lot better than i remembered, it's not fully gone so there is still something going on, maybe the rest could be the ping only updating once per second and maybe inconsistent frame rates

first half of the video it's off and second half it's on, this was in august
https://www.youtube.com/watch?v=1pgxVHjeOzM

OfficialLambdax:
Yeah the warping forth and back looked basically gone and the rest is likely explainable with packets not arriving for a considerable amount of time and frame inconcistencies as everything is still hooked to the frames. Thats nothing this feature can do anything about. all it does is to smooth the packet executions when they come through a unstable connection, because the VE lua cant deal with that case at all. But that only works well when you atleast got >40 fps

So all this pr here does is to bring this feature up to latest and thats it. I wont be touching it any further. The other issues have to be dealt with in manners i dont know, as such this pr can just be merged and this way be removed from the pr pile.

Olrosse (approved):

PR #487, opened 2024-01-08
[New Experimental] Increase FPS for decreased Damage Model

OfficialLambdax:
Full copyright goes to @Anonymous-275. Making it now a option in BeamMP after getting permission from them.

This piece of code only effects non player owned vehicles. I dont completly understand it, as i havent modded vehicles in this game yet, but i can see that the effects are reduced damage accuracy (like vehicles deform a bit differently), which yield more fps. I can join servers with more vehicles with this enabled. And i can see that players who dont care for damage accuracy will value this option.

Option `(Optimisation) Increase FPS for decreased Damage Model` is at default Off. Upon enabling it, it will affect all new spawned vehicles or those who have a edit applied to them. So the effect is not immediate if you are already in a server.


O1LER:
Where does shellNode get used? I see what the function does, it checks if a node is at the outside of a vehicle, but where does this get used then? If you say old code, was shellNode used in beamNG before collisionTriangles were a thing?

Basically it chooses for which node to override the jbeam values for collision and selfCollision, plus some aditional things to simplify tires. So my thoughts are, this could be refined with a distance check, with sliders like nametag visibility. If a local and a remote car are close to eachother, dont use the optimisation. If they are a certain distance away, use it. Might as well add a third one and disable collision/selfCollision on all nodes for very distant vehicles. Would be nice if we could just stop physics alltogether for very distant vehicles.

OfficialLambdax:
What i mean with old code is that this code was in a older ea version of beammp that was distributed through EA about 1 year ago. This piece of code, for a to me unknown reason, never made it into a actual release, even tho that it still works and that with the intended effect. Thats why i would like to launch it as a experimental

To your second suggestion. I lack total knowledge about vehicle modding in beamng, so it wont be me whos going to make that. but im totally with you. While further away collisions would not happen on your client, its still neat for a gain in frames. And this issue would be solved as soon as the vehicle state is synced.

Starystars67:
Hi @OfficialLambdax , Do you have any metrics to suggest the performance increase this offers?
Thanks

lionkor:
@OfficialLambdax

Unshown:
Enabled:
image
Disabled:
image

Results provided by discord user @thrustermaster

While at it i would like to ask for permission to test this code in a seperate client mod for real world testing and provide the results here unless this will get added to the EA Client in the next ~2 weeks.

thrustmaster10000:
Those results were achieved by injecting the logic into processNodes() stage of vehicle creation (stage2.lua). Vehicle creation after that does optimizations based on the nodes characteristics and their collision. like ordering (optimization.lua).

If modifying an already existing vehicle yields the same results, I don't know.

I did this because the code as it is now is incompatible with the physics benchmark.

```
local stage2 = require("vehicle/jbeam/stage2")

do
    local pushToPhysics = ({debug.getupvalue(stage2.loadVehicleStage2, 1)})[2]
    local processNodes  = ({debug.getupvalue(pushToPhysics, 1)})[2]
    debug.setupvalue(pushToPhysics, 1, function(vehicle)

*node processing*

    debug.setupvalue(pushToPhysics, 1, processNodes)
        return processNodes(vehicle)
    end)
end

return stage2
```

I feel the code would work well if implemented in this fashion instead, but I lack the knowledge on how to skip applying it to the player vehicle at that stage.

I'm also curious what the purpose of resetting physics is at the end.

Be aware this mod works differently to popular beamng optimization mods currently, (selfCollisionDisabler, luuk's performance mod) as it changes the node.collision values, which override all collision (even static). It keeps the outer shell nodes of the vehicle intact, and for wheels it skips every second node. I didn't notice much weirdness when testing this with traffic vehicles, everything looked fine and FPS was through the roof.

OfficialLambdax:
As this was integrated into a [performance specific mod](https://github.com/tmaster1000/node-performance-mod) i dont see a reason to keep this pr open

PR #488, opened 2024-01-08
fixed wrong body been used when simple traffic body doesnt exist

Please-Pick-a-Name:
before this, when the simple vehicle option is turned on, and if there is an entry in simplified_vehicles for the vehicle, it will attempt to replace the body slot of said vehicle to the simple traffic body
regardless if the part actually exist:
- instead of the original body, vehicles end up with the default body, if the part doesn't exist or not valid
added bonus:
since the recent beamng update, the naming convention for the simple traffic parts have changed, that means:
- all vehicles, include the one with the simple traffic variant, will have their body changed to default body

i want to fix it so it check if the part exists and the slot matches, and i want to update the list in simplified_vehicles

so far i have done the existence check, but idk how to test it, someone plz tell me beammp deving workflow

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -12,6 +12,7 @@
 
 local M = {}
 
+local jbeamIO = require('jbeam/io') -- to be used later for getting slotting information of parts
 
 -- ============= VARIABLES =============
 local lastResetTime = {}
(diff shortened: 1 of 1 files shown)


Please-Pick-a-Name:
it should now check if the part exist, and if the part's slot type matches/contains where it is expected to fit in
does not took into consideration of the new slots2 system, so it might fail to swap the part or swap the wrong part, but for now it will work since nobody really uses that system anyways
also updated the simplified_vehicles list
didnt test, dunno how to test

Please-Pick-a-Name:
fixed title cause i cannot english

Please-Pick-a-Name:
ik how to test my pr now...
yes it errored out, fixing rn

Please-Pick-a-Name:
actually figured out beammp contribute workflow, this "works_on_my_machine.txt"
thanks o1ler

Starystars67 (dismissed):
Testing this I had no errors or issues.
Thank you

Please-Pick-a-Name:
dead pr kewk

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -895,8 +905,16 @@ local function applyVehEdit(serverID, data)
 
 	if checkIfVehiclenameInvalid(vehicleName, playerName, vehicles[serverID]) then return end
 
-	if settings.getValue("simplifyRemoteVehicles") and simplified_vehicles[vehicleName] then
-		vehicleConfig.parts[vehicleName..'_body'] = simplified_vehicles[vehicleName]
+	if settings.getValue("simplifyRemoteVehicles") then
```
20dka, line 922: Could you extract the contents of the if check to a separate function to avoid duplicated code?

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -845,8 +847,16 @@ local function applyVehSpawn(event)
 
 	nextSpawnIsRemote = true -- this flag is used to indicate whether the next spawn is remote or not
 
-	if settings.getValue("simplifyRemoteVehicles") and simplified_vehicles[vehicleName] then
-		vehicleConfig.parts[vehicleName..'_body'] = simplified_vehicles[vehicleName]
+	if settings.getValue("simplifyRemoteVehicles") then
+		local expectedPartID=simplified_vehicles[vehicleName]
+		if expectedPartID then
+			local ioCtx={preloadedDirs={string.format("/vehicles/%s/",vehicleName),"/vehicles/common/"}} -- Fake io context for jbeamIO
+			local slotMap=jbeamIO.getAvailableSlotMap(ioCtx) -- slots2 compatible, maybe theres better way then require jbeamio and let it load the files again...
```
20dka:
```suggestion
			local slotMap = jbeamIO.getAvailableSlotMap(ioCtx) -- slots2 compatible, maybe theres better way then require jbeamio and let it load the files again...
```
Please leave whitespace around assignment statements. Do this everywhere in your code

20dka (changes requested):

lua/ge/extensions/MPVehicleGE.lua
Please-Pick-a-Name: sure, its now in vehicleGetSimplified instead

lua/ge/extensions/MPVehicleGE.lua
Please-Pick-a-Name: done

20dka (approved):

PR #509, opened 2024-01-26
Added stationary header for server list

### LINKED ISSUE #503: Stationary serverlist header
Serverlist header scrolls with the list. Would be better if it stayed visible at the top even when scrolling

AlexITA1100:
The search field and all the buttons are fixed at the top of the list.
The table head is a can of worms to keep fixed at the top for two reasons:
1. It's transparent; setting it to be `sticky` would solve the problem but all the other table rows will be visible below the `th` on scroll. This is why the rule is in the css but commented out.
2. If the `thead` or `th` are `fixed` the formatting of the table rows breaks spectacularly;

Due to these points I feel that this PR only partially resolves #503.


Starystars67 (dismissed):
Testing this indeed the header moves when you scroll, though I wonder how others think about the idea of this change?2

(Personally I like it)

Olrosse:
i don't know if this is new or if it's just my end but the mod size limiter slider is very tricky to use and i can only select 0 or 5000MB, nothing in between,

other than that it seems to work as expected :)

AlexITA1100:
> i don't know if this is new or if it's just my end but the mod size limiter slider is very tricky to use and i can only select 0 or 5000MB, nothing in between,
> 
> other than that it seems to work as expected :)

Oversight, fixed.

Starystars67 (approved):

Starystars67:
All checks out and appears to work as expected.

PR #514, opened 2024-02-08
Delete old LICENSE.md

OfficialLambdax:
Actual oversight, didnt notice that the new didnt overwrite the old. As such the old must be removed


lionkor (approved):
looks great!!!

PR #519, opened 2024-03-04
#304 Ability to restore deleted player vehicles

### LINKED ISSUE #304: [Feature] Way to restore deleted vehicles
As much as I love the ability to remove people that are trolling or are AFK, etc. I think there could be some improvements. Here is an example of what I find it quite annoying. If I remove someone because they are AFK and they begin playing right after, there is only one solution I have to fix this. I have to leave the server and come back. I know that the other player can fix it, but I would rather have a solution for undoing the removal I did. 

Now if it were a simple solution, it would be to add a UI button to add them back while tabbed onto them.

I would like for a way to toggle people on and off. I think that adding a toggle-able option to the player list UI would act as a great solution.

Josee-xav:
From my testing it seems to work perfectly. But if theirs any bugs i can fix it.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -727,9 +727,9 @@ function Vehicle:delete()
 	if players[self.ownerID] and self.serverVehicleString then players[self.ownerID].vehicles.IDs[self.serverVehicleString] = nil end
 	if self.serverVehicleString then vehicles[self.serverVehicleString] = nil end
 	
-    players_vehicle_configs[self.serverVehicleString] = nil
+    	players_vehicle_configs[self.serverVehicleString] = nil
 
-    self = nil
+    	self = nil
 end
 function Vehicle:onSerialized()
 	local t = {
@@ -1313,7 +1313,7 @@ local function onServerVehicleSpawned(playerRole, playerNickname, serverVehicleI
 			data = data
 		}
 
-        players_vehicle_configs[serverVehicleID] = decodedData
+        	players_vehicle_configs[serverVehicleID] = decodedData
 
 		if settings.getValue("enableSpawnQueue") and not (settings.getValue("queueSkipUnicycle") and decodedData.jbm == "unicycle") then
 			log("I", "onServerVehicleSpawned", "Adding spawn for " .. playerNickname .. " to queue")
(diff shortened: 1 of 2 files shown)


lionkor:
Hi, thank you for the contribution! Could you elaborate a bit on what the problem is that this fixes, etc. so it can be reviewed and tested easier?

lionkor:
Oh, it relates to #304 - nevermind my comment.

Starystars67:
I have tested this in-game and it does work as intended / expected.

Indeed the ui-revamp is going to be handling some of this aspect but in a slightly different way with how to allow the user to respawn their vehicle. 

In the meantime I will approve this as it does do as expected and there is currently no ETA on the UI-revamp.

Thank you for your contribution.

Starystars67 (dismissed):
Open to opinion, I think the function names should follow a Removed name rather than Destroyed. Happy to receive input on this otherwise approved ready to merge.

lionkor:
@20dka could we please get a review from you on the code semantics here?

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -719,7 +721,10 @@ function Vehicle:delete()
 	end
 	if players[self.ownerID] and self.serverVehicleString then players[self.ownerID].vehicles.IDs[self.serverVehicleString] = nil end
 	if self.serverVehicleString then vehicles[self.serverVehicleString] = nil end
-	self = nil
+	
+    players_vehicle_configs[self.serverVehicleString] = nil
```
lionkor: Would you mind changing the indentation to match the rest of the file? :)

lionkor: @clodiop could you please fix this and push that?

lionkor:
@20dka didn't review, so we'll move this forward now

lua/ge/extensions/MPVehicleGE.lua
Josee-xav: Should be fixed?

2PXL:
tested and it still works, thanks for making this.

Mack29446:
is this ready to merge?

Starystars67 (approved):

Starystars67:
I am going to go ahead and merge this at this time.

PR #528, opened 2024-03-20
Remove player list prefix and suffix in UI.lua

2PXL:


Starystars67:
Hello @2PXL, thank you for raising this PR.

Please could you provide a description and maybe before & after screenshot showing the change too?
I am fairly sure that I understand exactly what you are intending to do here but would be nice for it to be details for records.

Thanks

2PXL:
it removes everything that some servers add to the players nametag which can cause quite some clutter.
ba

2PXL:
forgot to mention that these changes do not affect the additional nametag stuff added by servers, just what shows up in the player list.
m

daniel-w0 (changes requested):
Looks good and functions as expected, just change these lines from:
```lua
username = username..''..player.role.shorttag
```

To:
```lua
username = username .. player.role.shorttag
```

As of right now, it's just concatenating nothing to username with `username = username .. ''`.

daniel-w0 (approved):
Thanks for the change. I've just tested it again and it works just fine :)

daniel-w0 (approved):
Looks and works just fine, thanks

Olrosse (approved):

PR #537, opened 2024-04-17
Fixed simplified vehicle part id again

Please-Pick-a-Name:
-- Added bx and removed coupe
-- Fixed wendover
-- Fixed bastion
-- Didn't fix fullsize
-- Didn't fix pickup
-- Didn't fix roamer
-- Didn't fix van


Please-Pick-a-Name:
We need a better implementation for this, we need to be able to customize which slot to replace instead of hardcode the body slot
I plan to making a new PR for that functionality, which could also select the correct simple body based on the current used parts so vehicle size/look match better, but i cant make it happen before the 21st

Starystars67 (approved):
Thank you for continuing to monitor this one.
I look forward to seeing your dynamic system for it later this month 👍

PR #545, opened 2024-05-01
Vehicle/Object Config Protection

### LINKED ISSUE #451: [Feature Request] Client support for Prevention of Vehicle Config Cloning and Configuration Saving
Some players do not want other players to be able to clone and / or save their vehicle configurations. This is reasonable, and support for it should be possible.

In BeamMP's Multiplayer Options, a pair of options for:
* Preventing Cloning, and
* Preventing Configuration Saving

If a player has either of these enabled, other players on the server should not be able to perform the corresponding action to that player's vehicle.

**Describe alternatives you've considered**
I've personally created a client mod that can prevent anyone from cloning or saving configurations, but this includes the player themself.

**Additional context**
Making other clients aware that a certain client prevents cloning or saving of the vehicle in question may be tricky, or involve server work to support it. This could potentially be a public resource of a server plugin / client mod pairing, but building this into BeamMP itself would be much more robust.

Starystars67:
The aim of this PR is to add code to make the saving and cloning of others vehicles harder. 

A user can enable the setting in the options to protect their configs.

From that point all of their vehicles spawned will be labelled as protected and then another user tries to clone or save it an error will be presented and the request denied.

Needs online testing with other players.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1132,8 +1132,8 @@ local core_vehicles_cloneCurrent = core_vehicles.cloneCurrent
 core_vehicles.cloneCurrent = function ()
 	local vehicle = be:getPlayerVehicle(0)
 	if vehicle:getField("protected", 0) == "1" then
-		local title = MPHelpers.translate("ui.multiplayer.configprotection.clone.title", "Vehicle Clone Error")
-		local msg = MPHelpers.translate("ui.multiplayer.configprotection.clone.message", "Sorry, you cannot clone this vehicle.")
+		local title = MPTranslate("ui.multiplayer.configprotection.clone.title", "Vehicle Clone Error")
+		local msg = MPTranslate("ui.multiplayer.configprotection.clone.message", "Sorry, you cannot clone this vehicle.")
 		guihooks.trigger("toastrMsg", {type="error", title=title, msg=msg})
 		return
 	else
@@ -1145,8 +1145,8 @@ local core_vehicle_partmgmt_saveLocal = extensions.core_vehicle_partmgmt.saveLoc
 local function core_vehicle_partmgmt_saveLocal_overwrite(p1)
 	local vehicle = be:getPlayerVehicle(0)
 	if vehicle:getField("protected", 0) == "1" then
-		local title = MPHelpers.translate("ui.multiplayer.configprotection.save.title", "Vehicle Save Error")
-		local msg = MPHelpers.translate("ui.multiplayer.configprotection.save.message", "Sorry, you cannot save this vehicle.")
+		local title = MPTranslate("ui.multiplayer.configprotection.save.title", "Vehicle Save Error")
+		local msg = MPTranslate("ui.multiplayer.configprotection.save.message", "Sorry, you cannot save this vehicle.")
 		guihooks.trigger("toastrMsg", {type="error", title=title, msg=msg})
 		return
 	else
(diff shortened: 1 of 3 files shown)


StanleyDudek:
As of https://github.com/BeamMP/BeamMP/pull/545/commits/325c861664d41980ef56497ba3fd28bc1010b871 this appears to be working for the clone current button, I tested two independent clients on two forum logins on the same official server to copy each

image

Same message presented to both when trying to use the clone current button


NOTE: configuration saving is **NOT** affected by this and a player can tab to someone's vehicle, then save it as a named configuration through the tuning menu, then spawn that vehicle. This is the same as cloning, but more steps, and we should see if we can prohibit this as well.

If I remember correctly, a difficulty here lies in a difference of access to different functions used by the clone current button and the configuration save button.

lionkor:
> This is the same as cloning, but more steps, and we should see if we can prohibit this as well.

agreed, could we see to get that fixed also?

Starystars67:
> configuration saving is **NOT** affected by this and a player can tab to someone's vehicle,

Looking over the code for how this works it calls partmgmt.saveLocal which is oveerwritten as part of this PR. 
See: https://github.com/BeamMP/BeamMP/pull/545/files#diff-0dea18c3fb9b9a9abe7baeef723ec9dbd98a148a5f3990947f084d1ab590fd42R998

Based on that this PR should address the case you mention. Do you have any suggestions as to why it does not?

StanleyDudek:
> Based on that this PR should address the case you mention. Do you have any suggestions as to why it does not?

To the best of my understanding, the html/js side of the save menu UI loads and holds the original save function from the base game, that while should be overridden as the pr indicates, it is not. Curious to note that this replacement approach does work for the clone current button, but, comparing the html/js of the vehicle select menu (where clone current button is) vs vehicle config menu (where save/load is via partial.save.html) that clone current is called from vehiclselect.js with `bngApi.engineLua('core_vehicles.cloneCurrent()', ...` and save vehicle is called from vehicleconfig.js with `bngApi.engineLua(`extensions.core_vehicle_partmgmt.saveLocal("${configName}.pc")`)`

I'm not entirely sure beyond this what a solution looks like.

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -928,7 +928,14 @@ local function sendVehicleSpawn(gameVehicleID)
 		vehicleTable.vcf = vehicleData.config -- Vehicle Config, contains paint data
 		vehicleTable.pos = {pos.x, pos.y, pos.z} -- Position
 		vehicleTable.rot = {rot.x, rot.y, rot.z, rot.w} -- Rotation
+		vehicleTable.pro = settings.getValue("protectConfigFromClone", false) -- Should the config be protected?
```
lionkor, line 1068: Could we name this properly so we don't need the comment? For example `protected`?

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -928,7 +928,14 @@ local function sendVehicleSpawn(gameVehicleID)
...
+		if vehicleTable.pro == true then
+			vehicleTable.pro = "1"
+		else
+			vehicleTable.pro = "0"
+		end
```
lionkor, line 1073: Is there a specific reason why we don't use a bool here?

Starystars67: The game gets funny with it sadly

lua/ge/extensions/MPVehicleGE.lua
Starystars67: I didnt want any issues to potentially arise in conflict with anything of the game or other details.

stefan750 (approved):

PR #547, opened 2024-05-01
Add Default settings

### LINKED ISSUE #460: [Bug] BeamMP causes a spam of "Unrecognized setting name "<setting here>" (not defined in defaults file: "/settings/defaults.json")" in the console when loaded or reloaded
**Describe the bug**
When BeamMP is loaded it causes a spam of the custom settings in the game console. This looks unpleasant and also adds more mess to the console window.

**To Reproduce**
Steps to reproduce the behaviour:
1. Load up BeamMP and check the console.

**Expected behaviour**
The game console should be clear and free of unnecessary output that serves no constructive output.

**Screenshots**
image

Starystars67:
This PR add's default settings and parses them the same way that we handle locales.
This reduces further spam in the console of the game:

image

The above is all now gone by this PR.

### DIFF
--- a/lua/ge/extensions/MPModManager.lua
+++ b/lua/ge/extensions/MPModManager.lua
@@ -22,14 +22,14 @@ function queueExtensionToLoad(extension)  -- temporary workaround for mods still
 end
 
 
-local function unloadLocales()
-	FS:unmount('/temp/beammp/beammp_locales.zip')
+local function unloadLocalesAndDefaults()
+	FS:unmount('/temp/beammp/beammp_locales_and_defaults.zip')
 	FS:directoryRemove('/temp/beammp')
 end
 
---- Load the BeamMP provided locales and merge them into a working set for BeamNG
-local function loadLocales() -- loads beammp locales without having to directly replace the game locales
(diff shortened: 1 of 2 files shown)


OfficialLambdax:
Just leaving this here; Default settings can be found in the MPConfig.lua https://github.com/BeamMP/BeamMP/blob/development/lua/ge/extensions/MPConfig.lua#L62

Starystars67:
@OfficialLambdax indeed the link you shared does hardcode some values there in Lua however what it does not handle is whether that setting is supposed to be cloud or locally saved. This PR addresses this issue which helps users maintain a consistent  experience across devices.

Olrosse:
does the game load these as well? reason I'm asking is because i put this into my second beam instance that i have changed to port 4446, and in the cache it shows the setting now as 4444 while in game it is still set to 4446

Starystars67:
> does the game load these as well? reason I'm asking is because i put this into my second beam instance that i have changed to port 4446, and in the cache it shows the setting now as 4444 while in game it is still set to 4446

These settings are just defaults, your configured option will still take precedence over it.

PR #548, opened 2024-05-02
reworked simplify vehicle functionality

Please-Pick-a-Name:
its now capable of finding the closest simple body depending on the given vehicle config parts

however the implementation starts turning into chaos dealing with vehicle like the pickup


Please-Pick-a-Name:
id like input on whatever can be done to make it better

lionkor:
is this ready for review? if so pls mark as ready

Please-Pick-a-Name:
> is this ready for review? if so pls mark as ready

sure i will mark this for review, since im not sure how i want to improve this mess further

Starystars67:
Testing this everything seems to work as expected for me.
I understand your concern about trying to tidy it up however I too am not sure on much that could be done at this point so I am going to move this through.

Starystars67 (approved):

PR #551, opened 2024-05-08
Changed LoadingServer tile to stay centered on all display aspect ratios

Mr-Rogue-Official:
Before vs after:
Screenshot 2024-05-08 213308
Screenshot 2024-05-08 213445
Additional SS for more context:
Screenshot 2024-05-08 213714

### DIFF
--- a/ui/modules/multiplayer/multiplayer.css
+++ b/ui/modules/multiplayer/multiplayer.css
@@ -128,6 +128,8 @@ md-content {
 }
 
 div#LoadingServer {
+  align-items: center;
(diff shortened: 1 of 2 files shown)


Starystars67:
Hi @Mr-Rogue-Official ,

~~Testing this myself I find that the tile is now lower on my screen than where it was previously.~~
~~It used to be upper middle where as now it is lower middle.~~

~~Otherwise, all seems to be good.~~

Edit: All good, seems to be me.

Starystars67 (approved):

PR #553, opened 2024-05-09
Fix for Hotlapping UI

### LINKED ISSUE #512: [Bug] Hotlapping UI app is broken when on a server
Yes

**Describe the bug**
When joined to a BeamMP server, the hotlapping UI app is missing its buttons, rendering it completely unusable.
image

**To Reproduce**
1. Join a BeamMP server
2. Press <kbd>ESC</kbd> and click `UI Apps`
3. Click `Add App`
4. Select the `Hotlapping` UI app
5. Drag the UI app from the top-left corner to somewhere you can see
6. Note that the buttons that are meant to be at the top of the UI app are no longer there. (see image below for how the app is supposed to look)

**Expected behavior**
I expect to be able to use the hotlapping UI app in multiplayer, as I can in singleplayer.
image

 - Windows 11 22H2

**Additional context**
I have performed a few tests to see where the issue stems from. Here are my findings:

1. `Safe Mode` and viewing the `Hotlapping` app is fine.
2. I tried normal and vulkan renderer, both seem to have no effect.
3. Launching BeamNG using the BeamMP launcher and entering freeroam is fine.
4. The UI app only breaks when launching using the BeamMP launcher and joining a server.

Starystars67:
So this way we can enable freeroam functionality properly without hindering the multiplayer UI layouts.

### DIFF
--- a/lua/ge/extensions/multiplayer/multiplayer.lua
+++ b/lua/ge/extensions/multiplayer/multiplayer.lua
@@ -121,7 +121,7 @@ local function onWorldReadyState(state)
 	if state == 2 then
 		if MPCoreNetwork and MPCoreNetwork.isMPSession() then
 			log('M', 'onWorldReadyState', 'Setting game state to multiplayer.')
-			core_gamestate.setGameState('multiplayer', 'multiplayer', 'multiplayer')
+			core_gamestate.setGameState('freeroam', 'multiplayer', 'multiplayer')
 			local spawnDefaultGroups = { "CameraSpawnPoints", "PlayerSpawnPoints", "PlayerDropPoints", "spawnpoints" }
 
 			for i, v in pairs(spawnDefaultGroups) do
--- a/lua/ge/extensions/multiplayer/multiplayer.lua
+++ b/lua/ge/extensions/multiplayer/multiplayer.lua
@@ -121,7 +121,7 @@ local function onWorldReadyState(state)
 	if state == 2 then
 		if MPCoreNetwork and MPCoreNetwork.isMPSession() then
 			log('M', 'onWorldReadyState', 'Setting game state to multiplayer.')
-			core_gamestate.setGameState('freeroam', 'multiplayer', 'multiplayer')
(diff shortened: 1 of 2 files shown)


StanleyDudek:
FYI:

`core_gamestate.setGameState('freeroam', 'multiplayer', 'multiplayer')`

This will make all freeroam available missions visible to all players at all times when on any server.

Starystars67:
Indeed this would occur however I am unable to come up with any other option other than the originally set out options.

Do you have any ides or suggestions?

Olrosse:
Ideally we would make a multiplayer version of the app and maybe with some tweaks for multiplayer, like have a setting where it only counts your own or the tabbed to car, i briefly tried this the other day but i don't understand how ui apps work so i didn't get very far

StanleyDudek:
I like the editing of `ui\modules\apps\Hotlapping\app.js`

I tested titch's proposed:
```js
        let reevaluateControls = function(){
             bngApi.engineLua('core_gamestate.state and core_gamestate.state.state', function(ret) {
              //console.log("Reevaluating on my own.. State = " + ret)
              scope.controlsEnabled = ret == 'freeroam' || ret == 'multiplayer'
            })
        }
```
this seems to work fine and other players do not seem to trip the lap / sections counter

StanleyDudek:
> I like the editing of `ui\modules\apps\Hotlapping\app.js`
> 
> I tested titch's proposed:
> 
> ```js
>         let reevaluateControls = function(){
>              bngApi.engineLua('core_gamestate.state and core_gamestate.state.state', function(ret) {
>               //console.log("Reevaluating on my own.. State = " + ret)
>               scope.controlsEnabled = ret == 'freeroam' || ret == 'multiplayer'
>             })
>         }
> ```
> 
> this seems to work fine and other players do not seem to trip the lap / sections counter

Also I did NOT test this tabbed to a vehicle but I think we can expect that there's no ownership check

lionkor:
Needs more discussion, as mentioned in dev meeting, this seems like it may not do what we want

PR #554, opened 2024-05-09
Added togglable player ID option.

### LINKED ISSUE #506: Remove player ID from playerlist
Playerlist constantly shows ID. Should be either a toggleable option, or only be shown on hovering the mouse over the list

Starystars67:
Addresses Issue #506

### DIFF
--- a/lua/ge/extensions/MPConfig.lua
+++ b/lua/ge/extensions/MPConfig.lua
@@ -60,7 +60,7 @@ local function setDefaultUnicycle(configFileName)
 end
 
 local defaultSettings = {
-	autoSyncVehicles = true, nameTagShowDistance = true, enableBlobs = true, showSpectators = true, nametagCharLimit = 32,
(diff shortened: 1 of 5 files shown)


OfficialLambdax (changes requested):
- [ ] When this setting is changed through the use of scripting, eg with `settings.setValue("showPlayerIDs", true)` then it wont effect the ui, until the ui is reloaded
- [x] The new config option is not part of the MPConfig.defaultSettings table. Default is now `false`

Starystars67:
> When this setting is changed through the use of scripting, eg with settings.setValue("showPlayerIDs", true) then it wont effect the ui, until the ui is reloaded

Yes, this will be because it does not call the relevant function to trigger it. See line 84-87 of `lua/ge/extensions/core/settings/settings.lua`. 

As such this is a user / scripter issue rather than an issue of BeamMP or the game.

Starystars67:
Regarding the default setting being missed, I would argue that it should be a part of PR #547 and should be added to that PR accordingly or to this one after #547 is merged.

PR #525, opened 2024-03-18
Fix paint sync + Fix Missing function for vehicleGetSimplified

Starystars67:
This PR fixes an issue currently in development whereby spawning a vehicle does not work due to vehicleGetSimplified not existing. I have renamed references to this to the correct function name of getVehicleSimplified.

This PR also fixes the paint syncing issue by essentially sending an edit request event message after confirming the vehicle to be local.

PR #557, opened 2024-05-13
Fixes Sync of vehicle colors by signaling the vehicle being ready when it actually is ready

OfficialLambdax:
## Problem
The `onExtensionLoaded` event in vehicle lua is called when this extension is loaded, its not a safe event to assume that when it is called the vehicle is actually ready. Tho BeamMP's code does so interpret it like that and tells BeamMP's GE lua that the vehicle is ready, which then leads to various issues when accessing the vehicle data, eg. the configs applied colors.

## Solution
Dont interpret `onExtensionLoaded` as a vehicle is ready signal, but use the `initLastStage` event. There might be safer events or solutions, but i couldnt find any documentation about the other events.

## Changes
1. Removed the `MPVehicleVE.onExtensionLoaded()` listener.
2. Added `MPVehicleVE.initLastStage()` listener
3. Undid the hacky color sync from this pr https://github.com/BeamMP/BeamMP/pull/525

## Ready for review
Tested with another member of the beammp team. Thanks to trebor8820
> Tested the following
1. Spawning a vanilla config with its default color
2. Spawning a vanilla config with a custom color
3. Cloning another players config
4. Spawning a entire custom config with its custom colors

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1291,9 +1291,6 @@ local function onServerVehicleSpawned(playerRole, playerNickname, serverVehicleI
 
 		log("W", "onServerVehicleSpawned", "ID is same as received ID, synced vehicle gameVehicleID: "..gameVehicleID.." with ServerID: "..serverVehicleID)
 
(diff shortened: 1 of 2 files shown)


Starystars67 (approved):
<3

PR #558, opened 2024-05-13
When using Color Picker apply paints as edit immediately

StanleyDudek:
to complement the new paint sync fix, please consider this, or let me know if you think there's a better way, but this is pretty good, imo

basically, when a player picks a color, gather up any necessary data to send a slimmed down edit packet to the server that specifically contains new paint data per paint slot, server and clients handle the rest and the paint comes through nice and fast without waiting for a 10 second edit auto sync, works fine with queue system, doesn't get spammy as far as I could tell

@OfficialLambdax

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1253,6 +1253,30 @@ local function onVehicleResetted(gameVehicleID)
 	end
 end
 
+--============================ ON VEHICLE COLOR CHANGED (CLIENT) ============================
+local function onVehicleColorChanged(gameVehicleID, index, paint)
+    if not MPCoreNetwork.isMPSession() then return end -- do nothing if singleplayer
+    local vehicle = getVehicleByGameID(gameVehicleID)
+    if vehicle and vehicle.serverVehicleString and vehicle.isLocal then -- If serverVehicleID not null and player own vehicle
+        local vehicleTable = {}
+        local vehicleData  = extensions.core_vehicle_manager.getVehicleData(gameVehicleID)
+        local veh          = be:getObjectByID(gameVehicleID)
+        local pos          = veh:getPosition()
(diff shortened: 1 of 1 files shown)


Starystars67:
I like the idea of this to be honest as this should be something that can be applied without any "lag" or delay on the end users from what I have seen.

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -1253,6 +1253,30 @@ local function onVehicleResetted(gameVehicleID)
...
+        local pos          = veh:getPosition()
+        local rot          = quat(veh:getRotation())
+        vehicleTable.pid = MPConfig.getPlayerServerID() -- Player Server ID
+        vehicleTable.jbm = veh:getJBeamFilename() -- JBeam
+        vehicleTable.pos = {pos.x, pos.y, pos.z} -- Position
+        vehicleTable.rot = {rot.x, rot.y, rot.z, rot.w} -- Rotation
+        vehicleTable.vcf = vehicleData.config -- Vehicle Config
+        for configData in pairs(vehicleTable.vcf) do --for all the data in the Config
+            if configData ~= "paints" then -- if the data is not paints
+                vehicleTable.vcf[configData] = nil -- remove it, we don't need it
```
OfficialLambdax, line 1273: We need it. A Oc packet (edit packet) overwrites the vehicle config and will break server side scripts. Just send a usual edit packet with all the data if for some reason the color changed

OfficialLambdax (changes requested):

lua/ge/extensions/MPVehicleGE.lua
StanleyDudek: Understood, going to close this PR, apply your requested changes, and roll this into another PR with other paint changes (reset color data) with a proper issue to go with it.

PR #559, opened 2024-05-13
Addresses an untested case from PR#557

OfficialLambdax:
https://github.com/BeamMP/BeamMP/pull/557

## Problem (Describtion by @StanleyDudek)
1. You spawn any car model with stock config
2. You replace your car to be same model but another CUSTOM config of that vehicle (should also have any custom color, but probably any color even stock)
3. For observers the color becomes something wrong but not always white or anything and not the previous color, it's like it has bad color data and picks something default

Eg.
1. Spawn a default blue pickup truck
2. Replace that pickup truck with another model but custom config, say custom red bx

## Solution (Idea by @StanleyDudek)
Read the color data directly from the vehicle object instead from the vehicle_manager.lua because it maybe not be ready yet when BeamMP reads it.

## Needs testing

### DIFF
--- a/lua/ge/extensions/MPHelpers.lua
+++ b/lua/ge/extensions/MPHelpers.lua
@@ -117,6 +117,58 @@ local function tableDiff(old, new)
 	return diff, o, n
 end
 
+--- Splits the given string by the given delimeter and optionally converts the slices into the desired type
+-- @param string string Some string to be split eg. "Hello World"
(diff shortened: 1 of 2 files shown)


StanleyDudek (approved):
This does what it says, the demonstrated case is resolved.

In addition, the new helper function is indeed helpful elsewhere for paints work.

Starystars67 (approved):
Tried and tested. Appears to do what it says on the tin.

PR #560, opened 2024-05-14
Addresses an untested case from PR#557, add color picker and traffic color sync

### LINKED ISSUE #270: [Bug] Traffic vehicle colors desync after short time
Yes

**Describe the bug**
Traffic vehicle colors desync after short time.

**To Reproduce**
1. At least two players on a server for observation.
2. Each player has a spawned vehicle. One player spawns at least one **traffic** vehicle on their end.
3. Vehicle begins with color synced between clients (color was sent on spawn).
4. After the traffic vehicle's first "respawn" (teleport out of line of sight, undo damage, change color, reset) vehicle color is desynced for player that did not spawn the traffic.

The Client sends a reset for the traffic vehicle, but the data sent on resets does not contain color data, and of course the server does not distribute that color data to clients because it does not exist (~~or the server may not send color data on resets even if it were present, as I was able to edit MPVehicleGE.lua to send color data during resets, but the server does not send it on~~).

**Expected behavior**
When a client's traffic vehicle resets, it should send color data to the server so the server can send it to clients.

 - OS + Version: Windows 11 2H22

**Additional context**
Doing this would also have the effect of allowing changed color data for player vehicles to sync on reset keypresses.
Traffic data may be even more involved, but this would be a start for a more 1:1 experience for players when using traffic in MP.

StanleyDudek:
This is a duplicate of https://github.com/BeamMP/BeamMP/pull/559

but it additionally supplies live color syncing via the color picker as well as syncing of traffic vehicles' colors

@OfficialLambdax did the work of adding the string to table splitter and getColorsFromVehObj() to MPHelpers

use of getColorsFromVehObj() within sendVehicleSpawn() sendVehicleEdit() and onVehicleColorChanged() will now allow us to ensure consistent vehicle colors at all times, and would close https://github.com/BeamMP/BeamMP/issues/270 

this PR removes a 0.2 second delay in onServerVehicleResetted()

My justification here is that edits don't have a delay and they're more intrusive.

If you think this would cause an issue let's talk about it.

The only piece after this is that when a player joins after someone has spawned a vehicle and chosen a new color, they will see the old color until the original player chooses a new vehicle, color, does any tuning or part edits, or simply hits the sync button. A server bug / feature requested would created for this.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1377,6 +1377,36 @@ local function onVehicleResetted(gameVehicleID)
 	end
 end
 
+--============================ ON VEHICLE COLOR CHANGED (CLIENT) ============================
+local function onVehicleColorChanged(gameVehicleID, index, paint)
+    if not MPCoreNetwork.isMPSession() then return end -- do nothing if singleplayer
+    local vehicle = getVehicleByGameID(gameVehicleID) -- get vehicle table for this vehicle
+    if vehicle and vehicle.serverVehicleString and vehicle.isLocal then -- If serverVehicleID not null and player own vehicle
+        local veh = be:getObjectByID(gameVehicleID) -- get vehicle as object
+		local pos = veh:getPosition() -- get position
+		local rot = quat(veh:getRotation()) -- get rotation
+		local vehicleTable = {
+			pos = {
+				x = pos.x,
+				y = pos.y,
+				z = pos.z
+			},
+			rot = {
+				x = rot.x,
+				y = rot.y,
+				z = rot.z,
+				w = rot.w
+			}
+		}
+		local vehicleData  = extensions.core_vehicle_manager.getVehicleData(gameVehicleID) -- get vehicle's data
+		vehicleTable.vcf = vehicleData.config -- put config in our table
+		vehicleTable.vcf.paints = MPHelpers.getColorsFromVehObj(veh) -- get paints
+        vehicleTable.vcf.paints[index] = paint --insert new paint at index as chosen from color picker
+		vehicleTable.pid = MPConfig.getPlayerServerID()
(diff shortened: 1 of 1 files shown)


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -1253,6 +1259,36 @@ local function onVehicleResetted(gameVehicleID)
...
+				w = rot.w
+			}
+		}
+		local vehicleData  = extensions.core_vehicle_manager.getVehicleData(gameVehicleID) -- get vehicle's data
+		vehicleTable.vcf = vehicleData.config -- put config in our table
+		vehicleTable.vcf.paints = MPHelpers.getColorsFromVehObj(veh) -- get paints
+        vehicleTable.vcf.paints[index] = paint --insert new paint at index as chosen from color picker
+		vehicleTable.pid = MPConfig.getPlayerServerID()
+		vehicleTable.jbm = veh:getJBeamFilename()
+		MPGameNetwork.send('Or:'..vehicle.serverVehicleString..":"..jsonEncode(vehicleTable).."")
```
OfficialLambdax, line 1407: Im blocking this pr to have it brought up in a dev meeting, as it changes the protocol and also requires changes in the server

OfficialLambdax: Topic has been brought forward. The necessary changes will be done by @lionkor. Afterwards this pr can moved forward

Starystars67: @lionkor @OfficialLambdax any further updates to this?
E.g. is there a server version that this has been ear marked for yet?

OfficialLambdax: > is there a server version that this has been ear marked for yet?

Not that i know of

20dka:
we can revisit this in the near future, as a new feature

PR #561, opened 2024-05-15
revised input syncing and added support for custom inputs

Olrosse:
pretty much a rewrite of the input syncing code to support custom inputs and to make steering more responsive

changed steering input filter to direct and instead added temporal smoothing to every input, seems a lot more accurate and more responsive than using the pad filter, it does however add a bit of latency to inputs that where already using direct filtering

disabled local inputs on the remote vehicles so ghost controlling is no longer possible

changed gear sync to send the current gear with any input change so if a queue is spawned late it will sync as soon as there is any input

### DIFF
--- a/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
+++ b/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
@@ -83,7 +83,7 @@ local function applyGear(data) --TODO: add handling for mismatched gearbox types
 end
 
 local function getInputs()
-	local inputTableToSend = {}
+	local inputsToSend = {}
 	for inputName, _ in pairs(input.state) do
 		local state = electrics.values[inputName] -- the electric is the most accurate place to get the input value, the state.val is different with different filters and using the smoother states causes wrong inputs in arcade mode
 		if state then
@@ -97,18 +97,13 @@ local function getInputs()
 			end
 			state = math.floor(state * 1000) / 1000
 			if lastInputsTable[inputName] ~= state then
-				inputTableToSend[inputName] = {state = state}
+				inputsToSend[inputName] = {state = state}
 				lastInputsTable[inputName] = state
 			end
 		end
 	end
-	if not tableIsEmpty(inputTableToSend) then
-		inputTableToSend.g = electrics.values.gear -- if there is any input we also send the gear in case a remote vehicle is spawned after it has been put into gear
-		obj:queueGameEngineLua("MPInputsGE.sendInputs(\'"..jsonEncode(inputTableToSend).."\', "..obj:getID()..")") -- Send it to GE lua
-	end
 
 	 -- Old sync, keeping for temporary cross compatibility --
-	local inputsToSend = {}
 	currentInputs = {
 		s = electrics.values.steering_input and math.floor(electrics.values.steering_input * 1000) / 1000,
 		t = electrics.values.throttle and math.floor(electrics.values.throttle * 100) / 100,
--- a/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
+++ b/lua/vehicle/extensions/BeamMP/MPInputsVE.lua
@@ -5,7 +5,6 @@
 local M = {}
 
 -- ============= VARIABLES =============
-local currentInputs = {}
 local lastInputs = {
 	s = 0,
 	t = 0,
@@ -14,9 +13,9 @@ local lastInputs = {
 	c = 0,
 }
 
-local lastInputsTable = {}
 local inputCache = {}
 
+local periodicGearSyncTimer = 0
 local remoteGear
 local unsupportedPowertrainDevice = false
 local unsupportedPowertrainGearbox = false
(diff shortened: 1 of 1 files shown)


lua/vehicle/extensions/BeamMP/MPInputsVE.lua
```diff
@@ -78,13 +82,28 @@ local function applyGear(data) --TODO: add handling for mismatched gearbox types
...
+				if v.data.input then
+					state = -state / (v.data.input.steeringWheelLock or 1) -- converts steering wheel degrees to an input value
+				end
+			end
+			if math.abs(state) < 0.001 then -- prevent super small values to count as updates
+				state = 0
+			end
+			state = math.floor(state * 1000) / 1000
+			if lastInputsTable[inputName] ~= state then
+				inputsToSend[inputName] = {state = state}
```
OfficialLambdax: is this nesting necessary?
eg. `"steering":{"state":-0.172}`. why not just `"s":-0.172`?

lua/vehicle/extensions/BeamMP/MPInputsVE.lua
```diff
@@ -78,13 +82,28 @@ local function applyGear(data) --TODO: add handling for mismatched gearbox types
...
+				lastInputsTable[inputName] = state
+			end
+		end
+	end
+
+	 -- Old sync, keeping for temporary cross compatibility --
```
OfficialLambdax: How temporary? im thinking of just disabling/removing the old code and make the new match the old format

lua/vehicle/extensions/BeamMP/MPInputsVE.lua
```diff
@@ -100,26 +119,103 @@ local function getInputs()
...
+end
+
+local function onExtensionLoaded()
+	for inputName, _ in pairs(input.state) do
+		if not inputCache[inputName] then
+			inputCache[inputName] = {
+				smoother = newTemporalSmoothing(1, 1, nil, 0),
+				currentValue = 0,
+				state = 0,
+				diffrence = 0
```
OfficialLambdax: This typo in `diffrence` instead of `difference` is used multiple times in this pr, tho in `onReset()` this key is named `inputData.difference`

OfficialLambdax (changes requested):
## Problem
About `if a queue is spawned late it will sync as soon as there is any input`
Issue is this
1. Remote player spawned vehicle (i kept it in the quo)
2. They changed to second gear and pressed the clutch
3. I quod
4. I tabbed to their vehicle and could see the gear being in the second but the vehicle was stalled, no clutch input was applied

## This pr changes the packet protocol
### Current format
```
  [c] = float
  [g] = int/string
  [t] = float
  [s] = float
  [b] = float
  [p] = float
```
It only sends each key over this protocol when there has been any update to it

### This PR's format
```
  [throtte] = table
    [state] = float
  [brake] = table
    [state] = float
  [steering] = table
    [state] = float
  [clutch] = table
    [state] = float
  [parkingbrake] = table
    [state] = float
```
The transmitted contents are similiar to the old, but it keeps sending the gear index with every inputs update even when the gear has not changed

## Summary
I think the new system should match the protocol of the old as in the packets contents, aka "steering" gets translated to "s" to reduce packet size. Otherwise the pr appeared to be working as intended, even if it does mix the new with the old protocol. Even testing this with a player that had a high ping to me (~150ms) the inputs update was fluid.

lua/vehicle/extensions/BeamMP/MPInputsVE.lua
Olrosse: no not really, the only reason it's there is because i first included the filter as well and after removing that it was convenient to just check for table on receiving end instead of checking for every of the old names, but i can change it to just using the old names, though any custom input from mods will of course have to use their actual input name cause we would have no idea what the input name would be

lua/vehicle/extensions/BeamMP/MPInputsVE.lua
Olrosse: yep can do that, it's only there so it sync across this build and the current public build, i was gonna remove it as soon as this is in public

OfficialLambdax: Yes i see tho as soon as this pr is merged and a new release is build, it would automatically be shipped to everyone. So the cross compatibility is really not necessary.

lua/vehicle/extensions/BeamMP/MPInputsVE.lua
OfficialLambdax: `though any custom input from mods will of course have to use their actual input name cause we would have no idea what the input name would be` yes so we only want to translate the common inputs, eg steering, braking etc. Really clever idea btw to just attach custom inputs

OfficialLambdax:
### Todo
- [x] Translate common inputs to shortend versions eg. `steering` -> `s`
- [x] Remove nesting eg. `"s":{state:0.111}` ->`"s":0.111`
- [x] Remove cross compatibility code
- [x] Fix typo
- [x] I have to test custom input mods. Are there any that come to mind?
- [x] I have to test gear grind syncing

OfficialLambdax (commented):
PR and Changes work as expected, thank you @Olrosse and @stefan750.

Locked for: Will perform a more sorrow test within a competitive community to see how well it runs given varying ping and game lag. Just to see if this pr makes it worse.

lionkor:
Could we possibly get this tested without an event, or just earlier in general?

Starystars67:
@OfficialLambdax any update on this?
Thanks

Starystars67 (approved):

PR #565, opened 2024-06-03
fix for VE extensions not loading with some vehicle configs

Olrosse:
some configs like the Legran derby configs calls onVehicleSpawned twice for some reason, Which is another bug to look at, But this is causing an edge case where the lua thinks it's an edit instead of a spawn which means VE extensions don't get loaded,
I've fixed the extensions not loading by moving the loadModulesInDirectory queue function out of the spawn/edit checks so it runs every time onVehicleSpawned is called,

This does not cause duplicate extensions from my testing so it shouldn't cause any new issues, I tested it by making a local counter count up every frame in a VE file to see if it ever repeated a number after edits or after spamming the load function in the console, it never duplicated a single number

I also removed addModulePath cause it doesn't seem to be needed, If there is a reason it's there then let me know and i will add it back, the game also doesn't use it when loading the powertrain, controller, energyStorage and auto extensions which makes me think it's not necessary

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -999,9 +999,6 @@ local function onVehicleSpawned(gameVehicleID)
 		log("I", "onVehicleSpawned", "New Vehicle Spawned "..gameVehicleID)
 		--log('E', 'nextSpawnIsRemote', tostring(nextSpawnIsRemote))
 
-		veh:queueLuaCommand("extensions.addModulePath('lua/vehicle/extensions/BeamMP')") -- Load lua files
-		veh:queueLuaCommand("extensions.loadModulesInDirectory('lua/vehicle/extensions/BeamMP')")
-
 		if not nextSpawnIsRemote then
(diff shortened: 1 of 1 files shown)


Starystars67:
Testing this I did not find any obvious issues, BeamMP VE files are indeed still loaded.

Thank you

Starystars67 (approved):

Starystars67:
I would appreciate either @OfficialLambdax or @20dka 's approval on this as well though before merging.

OfficialLambdax (approved):
This is a trivial change that will likely have zero negative effects. Except it being "unclean".
`extensions.loadModulesInDirection()` eventually just calls `require()` which by lua standards wont reload a file if its already present within the `package.packages` table.

Tho it is described that a lua file can be entirely reloaded with `require` if the package is removed first, tho i dont see the extensions lua from `BeamNG.drive\lua\common\extensions.lua` doing that.

Will not cross test, since @Starystars67 already did it.

Starystars67:
For lack of a better option currently I will continue and merge this.

PR #566, opened 2024-06-04
Complete warning prompt when joining servers with mods.

### LINKED ISSUE #563: [Feature Request] Integrate Mod download Confirm button thats shown on join of a modded server
1. Upon the launcher connecting to the Server and it receiving the modlist of the server (during sync) the launcher shall give the game client a signal and then wait.
2. The BeamMP mod should then display a confirmation box to the player, letting them choose to actually join this server or aboard it. The reponse of that should be given back to the launcher.
3. On Yes, the launcher will continue as is with the sync protocoll
4. On No, the launcher will quit the joining process

Starystars67:
This PR adds new functionality to prompt the user about the downloading of mods and that they could contain malicious code giving the user the choice on if they want to take that risk or not.

image

**Before merging please ensure that all translations are completed.**


lionkor:
Notes from the call: Add text that says "you can configure this in the servers" and add configuration option(s) for it

PR #570, opened 2024-06-17
Paint checks and reverted onVehicleReady change

Olrosse:
Non color able vehicles like the unicycle snowman and most props caused the spawn packet to have empty baseColor because getField only returned "white", This caused the vehicle to fail to spawn on receiving end, I've just made it check if paint data exists and if not fill it in with basic paint data.

onVehicleReady didn't get called because initLastStage doesn't get called for extensions and is called before the BeamMP extensions are loaded, this caused Key Event Listeners to not get loaded and GE to not set remote vehicles to remote on vehicle edits, I've just reverted it to onExtensionLoaded because our extensions are loaded after the vehicle is ready anyways

### DIFF
--- a/lua/ge/extensions/MPHelpers.lua
+++ b/lua/ge/extensions/MPHelpers.lua
@@ -159,12 +159,21 @@ local function getColorsFromVehObj(veh)
 	local paints = {}
 	paints[1] = translate_metallic(splitStringToTable(veh:getField("metallicPaintData", 0), " ", 1))
 	paints[1].baseColor = splitStringToTable(veh:getField("color", 0), " ", 1)
-	
+	if tableIsEmpty(paints[1].baseColor) then -- making sure baseColor isn't empty causing failed spawn on recieving end
+		paints[1].baseColor = {1,1,1,1}
+	end
+
 	paints[2] = translate_metallic(splitStringToTable(veh:getField("metallicPaintData", 1), " ", 1))
 	paints[2].baseColor = splitStringToTable(veh:getField("colorPalette0", 0), " ", 1)
-	
+	if tableIsEmpty(paints[2].baseColor) then
+		paints[2].baseColor = {1,1,1,1}
+	end
+
 	paints[3] = translate_metallic(splitStringToTable(veh:getField("metallicPaintData", 2), " ", 1))
 	paints[3].baseColor = splitStringToTable(veh:getField("colorPalette1", 0), " ", 1)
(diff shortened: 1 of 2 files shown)


OfficialLambdax:
I feel this maybe a thing of *"some systems behave like this and others like that"*. cause we tested #557 across systems before merging. And the effect it had was to basically fix the color sync, except the later discovered case in #559

Starystars67:
> I feel this maybe a thing of _"some systems behave like this and others like that"_. cause we tested #557 across systems before merging. And the effect it had was to basically fix the color sync, except the later discovered case in #559

So what is the way forward here then peeps?

OfficialLambdax:
> So what is the way forward here then peeps?

the `initLastStage` approach ran well on any systems that helped testing it, except on olrosses. Idk the cause, but reverting back to onVehicleReady sounds like the only option until we know why it happens

OfficialLambdax:
Tests
- [x] Spawn a vanilla config with its default color
- [x] Spawn a vanilla config with a custom preconfigured color
- [x] Spawn a vanilla config with a entirely custom color
- [x] Edit into a vanilla config with its default color
- [x] Edit into a vanilla config with a custom preconfigured color
- [x] Edit into a vanilla config with a entirely custom color
- [x] Spawn a snowman unicycle
- [x] Spawn a custom config with its default color
- [ ] Spawn a custom config with a custom preconfigured color
> cannot test, preconfigure color is not applied on the spawning client
- [ ] Spawn a custom config with a entirely custom color
> cannot test, preconfigure color is not applied on the spawning client
- [x] Edit into a custom config with its default color
- [ ] Edit into a custom config with a custom preconfigured color
> cannot test, preconfigure color is not applied on the spawning client
- [ ] Edit into a custom config with a entirely custom color
> cannot test, preconfigure color is not applied on the spawning client
- [ ] 1) Spawn vanilla config with default color 2) Edit to the same vehicle but with a custom config
> Some times failes

Additional Notes
- Reading the color palette for the base colors from a vehicle object can return strings as eg. "InvinsibleBlack"

OfficialLambdax (approved):
1. Remote vehicles are correctly asumed to be remote vehicles upon spawn and edit
2. Color Sync is still borked in one case that this pr doesnt cover, see https://github.com/BeamMP/BeamMP/issues/572.

Test cases as stated in the comment above.

Approving as the last color sync issue needs a more in depth look, which is likely touched in a different pr

Tested with
- @Starystars67

Starystars67 (approved):

PR #573, opened 2024-06-23
Implement binary header for launcher to game communication to allow for binary packets

### LINKED ISSUE #443: [Bug] Double vehicle spawn and unsync
Yes

**Describe the bug**
When I join a server for the first time I get a lua error after spawning my vehicle additionally a second vehicle spawns, and I am not synced to anybody else, I can see everyone move and do their life but no one can see me move, other player only see one vehicle

**Screenshots**
image

 - OS + Version: Windows 11

**Additional context**
Once the double vehicles are spawned I cannot spawn or replace them, my camera just moves to someone else vehicle

This does not happen after I quit the server and join again

### LINKED ISSUE #483: [Bug] Stop using newline as a packet separator
Currently, we read line-by-line from the socket. This is not appropriate behavior. The right way to read from a TCP socket is to first send a fixed-length prefix which contains the length of the following packet.
Newlines `\n` (0x0a) can appear in *any* kind of data, this is a source of bugs (and exploitable in a pretty bad way, afaik).

I propose the fix of sending either

1. A 4-byte binary prefix and reading it with luajit ffi (@snepsnepsnep has an implementation of this and can likely make one for this)
2. An 8-byte zero-padded hexadecimal string like `00002364`, can be read with standard lua functions iirc

As a prefix for *each* message from the launcher to the lua. The same should be implemented for the lua to the game, for consistency.

lionkor:
No more need for `\n`.

### DIFF
--- a/lua/ge/extensions/MPGameNetwork.lua
+++ b/lua/ge/extensions/MPGameNetwork.lua
@@ -72,7 +72,7 @@ end
 --- Send given packet to the Server, over the Launcher.
 -- @tparam string s The packet to be sent to the Launcher/server.
 -- @usage MPGameNetwork.sendData(`<data>`)
-local function sendData(data)
+local function sendData(data) -- TODO currently the socket keeps retrying indefinitely if timed out, this freezes the game if the launcher is frozen, breaking the loop with offset the header and break the connection, we could maybe buffer data and try again next frame?
 	-- if not connected return
 	if not TCPLauncherSocket then return end
 	local header = ffi.string(ffi.new("uint32_t[?]", 4, #data), 4)
--- a/lua/ge/extensions/MPGameNetwork.lua
+++ b/lua/ge/extensions/MPGameNetwork.lua
@@ -473,6 +473,15 @@ local HandleNetwork = {
 
 local heartbeatTimer = 0
 
+local recvState = {
+	-- 'ready': ready to receive a new packet, data is contained within `data` if any
+	-- 'partial': `partialData` contains data, we're missing `missing` bytes
+	-- 'error': errorneous state
+	state = 'ready',
+	data = "",
+	missing = 0,
+}
+
 --- Tries to receive data from the Launcher every tick from the gameengine and handles the launcher <-> game heartbeat.
 -- @tparam integer dt delta time
 -- @usage INTERNAL ONLY / GAME SPECIFIC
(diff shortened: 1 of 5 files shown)


lionkor:
Related to https://github.com/BeamMP/BeamMP/pull/479

lionkor:
Explanation:

We implemented a new packet format for the launcher-to-game communication. The old format, where `<message>` is the message content, was:
```
<message>\n
```
The new format:
```
size<message>
```
Here `size` is 4 bytes of binary data, containing a little-endian unsigned 32-bit integer. The value is equal to the length of the `<message>` in bytes.

In order to handle some error and edge cases, we keep a state called `recvState` across multiple calls to `receive`. The `receive` function is effectively a state machine with three states; "ready", "partial", and "error".

We use these states to handle two edge cases:

1. Since the timeout is zero, it's technically possible that the transmission does not finish(?), at which point only a partial packet is transmitted. We observe that the returned status is `'timeout'`, that's how we can tell. We require that at least the 4 bytes for the header are transmitted successfully, we consider anything else an error state. When the 4 bytes are read, we then `receive()` the specified number of bytes. If we receive a number of bytes less than we expected, we enter the `partial` state. We can enter a `partial` state from either the `ready` or the `partial` state (the latter because we can *technically* keep receiving small bits of partial data).
2. If `client:receive(...)` fails with `'error'`, we enter the `error` state, which we can't leave by receiving more.

We parse the size by use of `ffi`. We assume that this code is slower than the previous implementation, simply because of the amount of code ran every time, and the `ffi` call. However, we did not notice any performance difference, and this is important, so we ignored that.

lionkor:
We still need to build the same for the other side of the transmission, as we encountered weird issues when sending large node data for vehicles with many nodes, and the launcher of course can't handle partial receives at the moment.

This can be reviewed and merged regardless, as it's technically a feature that works in isolation (with the launcher from https://github.com/BeamMP/BeamMP-Launcher/pull/98)

lua/ge/extensions/MPGameNetwork.lua
```diff
@@ -74,21 +62,29 @@ end
...
+		while (retries > 0 and error) do
+			isConnecting = false
+			log('E', 'sendData', 'Socket error: '..error)
+			if error == "timeout" then
+				log('W', 'sendData', 'Stopped at index: '..index..' while trying to send '..#packet..' bytes of data. retries:' .. retries)
+				packet = string.sub(packet, index + 1)
+
+				bytes, error, index = TCPLauncherSocket:send(packet)
+			end
 		end
```
lionkor, line 95: this is not complete, we need to handle the error case(s)

lionkor:
I've been using this exclusively for the last few days, so a few hours of play in total, with the corresponding launcher version, and I have not seen any issues. I might have missed something.

Let's please all run and test this version for a 1+ hour(s) each @Starystars67 @OfficialLambdax @snepsnepsnep @20dka and whoever else can. The Launcher builds are here (or build it yourself from the launcher's `better-header` branch):

[BeamMP-Launcher.exe.zip](https://github.com/user-attachments/files/16025896/BeamMP-Launcher.exe.10.zip)

or grab it from the [workflow](https://github.com/BeamMP/BeamMP-Launcher/actions/runs/9709067032/artifacts/1647701502) if you have access.

Starystars67:
LOCKED pending Launcher Update 2.2.

OfficialLambdax:
Would likely fix the rare #443 bug

Olrosse:
This should be ready now, partial send could be handled better and is currently blocking, Which could cause performance drop when sending a lot of data, but i would not deem this critical and should in theory never be happening.

just a reminder it will require the better header redone launcher to connect, so that will have to be released with this : https://github.com/BeamMP/BeamMP-Launcher/pull/139

Olrosse (approved):

lua/ge/extensions/MPGameNetwork.lua
Olrosse: could we just break the loop if the error is not timeout and let the code below handle the other errors? technically only timeout would need a loop, right?,

But i think a change like that needs proper testing, so maybe we should look at this later, the mod seems to function correctly as it is now so i don't think it's worth changing this right now

Issue #572, opened 2024-06-20
[Bug] MPHelpers.getColorsFromVehObj() fails to read colorPaletteX when baseColor in Config is [0,0,0,0]


OfficialLambdax:
**Describe the bug**
Should a player choose to spawn a custom config that contains any base color like this in .paints
```
"baseColor":[
        0,
        0,
        0,
        0
],
```
`veh:getField("colorPalette0", 0)` in MPHelpers.getColorsFromVehObj() will return the string `InvinsibleBlack` instead of the base color as string which leads to BeamMP replace it with [1,1,1,1] therefore desyncing the actual color of the vehicle.

**To Reproduce**
1. Create a custom config, where one of the basecolors is [0,0,0,0]
2. Spawn the vehicle
3. Open lua console
4. run `print(be:getPlayerVehicle(0):getField("colorPalette0", 0))` in GE Lua

OfficialLambdax:
Example config
[example.zip](https://github.com/user-attachments/files/15916244/example.zip)

OfficialLambdax:
Fixed with https://github.com/BeamMP/BeamMP/pull/576

PR #576, opened 2024-07-03
changed the way we get the colors on vehicle spawn

Olrosse:
using getField had issues where the base color could return as nil, this should hopefully fix that by using getMetallicPaintData and veh.color, veh.colorPalette etc instead.

I also removed a duplicate of getObjectByID

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -923,10 +923,7 @@ local function sendVehicleSpawn(gameVehicleID)
 		vehicleTable.rot = {rot.x, rot.y, rot.z, rot.w} -- Rotation
 		
 		-- The vehicle_manager.lua may not contain the correct color values, since v0.31, when we read them from that lua, so we read those from the object itself
-		local metallicPaintData = veh:getMetallicPaintData()
-		vehicleTable.vcf.paints[1] = createVehiclePaint(veh.color, metallicPaintData[1])
-		vehicleTable.vcf.paints[2] = createVehiclePaint(veh.colorPalette0, metallicPaintData[2])
-		vehicleTable.vcf.paints[3] = createVehiclePaint(veh.colorPalette1, metallicPaintData[3])
+		vehicleTable.vcf.paints = MPHelpers.getColorsFromVehObj(veh)
 
 		local stringToSend = jsonEncode(vehicleTable) -- Encode table to send it as json string
 		MPGameNetwork.send('Os:0:'..stringToSend) -- Send table that contain all vehicle informations for each vehicle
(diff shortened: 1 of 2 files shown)


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -924,7 +923,10 @@ local function sendVehicleSpawn(gameVehicleID)
 		vehicleTable.rot = {rot.x, rot.y, rot.z, rot.w} -- Rotation
 		
 		-- The vehicle_manager.lua may not contain the correct color values, since v0.31, when we read them from that lua, so we read those from the object itself
-		vehicleTable.vcf.paints = MPHelpers.getColorsFromVehObj(veh)
+		local metallicPaintData = veh:getMetallicPaintData()
+		vehicleTable.vcf.paints[1] = createVehiclePaint(veh.color, metallicPaintData[1])
+		vehicleTable.vcf.paints[2] = createVehiclePaint(veh.colorPalette0, metallicPaintData[2])
+		vehicleTable.vcf.paints[3] = createVehiclePaint(veh.colorPalette1, metallicPaintData[3])
```
OfficialLambdax: I think this should be refactored into the `MPHelpers.getColorsFromVehObj()` function. Point is that atleast one more pr is building on top of it #560

OfficialLambdax (changes requested):

lua/ge/extensions/MPVehicleGE.lua
Olrosse: I'll change it once I'm on, i was originally thinking we could just do the same for that PR as well but yeah having it as one function makes more sense

OfficialLambdax:
Tests
- [x] Spawn a vanilla config with its default color
- [x] Spawn a vanilla config with a custom preconfigured color
- [x] Spawn a vanilla config with a entirely custom color
- [x] Edit into a vanilla config with its default color
- [x] Edit into a vanilla config with a custom preconfigured color
- [x] Edit into a vanilla config with a entirely custom color
- [x] Spawn a snowman unicycle
- [x] Spawn a custom config with its default color
- [ ] Spawn a custom config with a custom preconfigured color
- [ ] Spawn a custom config with a entirely custom color
- [x] Edit into a custom config with its default color
- [ ] Edit into a custom config with a custom preconfigured color
- [ ] Edit into a custom config with a entirely custom color
- [x] 1) Spawn vanilla config with default color 2) Edit to the same vehicle but with a custom config
- [x] Spawn the config from this issue https://github.com/BeamMP/BeamMP/issues/572

SaltySnail:
The unmarked tests don't work in vanilla beamng for certain configs, but the colors are the same between clients.

OfficialLambdax (approved):
Approving and merging on the basis of the test results from the two dev team members. Can confirm that the missing checkmarks are based on game issues.

Thanks @Olrosse for the pr. Thanks @SaltySnail for helping in testin (:

PR #579, opened 2024-07-05
centred nametags and fixed spectator nametags not being cleared

### LINKED ISSUE #352: [Bug] Spectatorship does not get updated on VehicleSpawn
**Describe the bug**
When you are spectating someones Vehicle and then Spawn/Clone a vehicle, while watching them, then your focus will be Switched to your freshly Spawned vehicle. BeamMP does however not update the Spectatorship to the Server. This is why you can see Players Spectating someone else, while they are driving their own vehicle.

This may be required to fix for the implementation of this
https://github.com/BeamMP/BeamMP-Server/issues/184

I can tell with certanity that the according Spectator Change packet (Om) is not send. Because it does not arrive in my bot.

**To Reproduce**
1. Join a Server with Players in it with 2 Clients (1 you, 1 somone that is helping you)
2. Switch to a Vehicle that is not yours and not of the second client
3. Spawn a new Vehicle or clone that vehicle
4. With the Second Client look at the Vehicle you where spectating. You will see that you are still listed as spectating them (while you are in fact not).

**Expected behavior**
Switch the Spectator to the Fresh spawned vehicle

Olrosse:
Currently the name tags are placed 2 meters above the ref node for every vehicle, Here I've moved them to the center of the vehicle 20 cm above the top of the bounding box.
#
Before
screenshot_2024-07-04_22-16-35
After
screenshot_2024-07-04_22-15-21
#
I've also fixed the bug where spectator tags would stay if you cloned or spawned a car while being tabbed to someone else, this happened because the onVehicleSwitched function runs before the vehicle exists in the vehicle map, to fix this i moved the code that sends the active vehicle into it's own function, And then call it when the car is added to the vehicle map if you are tabbed to it.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1967,7 +1967,7 @@ local function onPreRender(dt)
 					v.vehicleHeight = veh:getInitialHeight()
 				end
 				local tempPosx,tempPosy,tempPosz = be:getObjectOOBBCenterXYZ(gameVehicleID)
-        		v.position = vec3(tempPosx,tempPosy,tempPosz)
+				v.position = vec3(tempPosx,tempPosy,tempPosz)
 				v.position.z = v.position.z + (v.vehicleHeight * 0.5) + 0.2
 			end
 
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1540,9 +1540,11 @@ local function onServerVehicleRemoved(serverVehicleID)
 			--vehicle:delete()
 		else
 			log('W', "onServerVehicleRemoved", "Failed removing vehicle "..serverVehicleID..", Vehicle can't be found")
+			vehicle:delete()
 		end
 	else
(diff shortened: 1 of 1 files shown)


Starystars67:
Woah!

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -1945,7 +1963,12 @@ local function onPreRender(dt)
 			local veh = be:getObjectByID(gameVehicleID)
 
 			if v.isSpawned and veh then -- update position if available
-				v.position = veh:getPosition()
+				if not v.vehicleHeight then
+					v.vehicleHeight = veh:getInitialHeight()
+				end
+				local tempPosx,tempPosy,tempPosz = be:getObjectOOBBCenterXYZ(gameVehicleID)
+        		v.position = vec3(tempPosx,tempPosy,tempPosz)
```
Starystars67: Indenting mistake :P

Starystars67 (changes requested):

Starystars67 (approved):
Pending a final test this is good to go!

WiserTixx (approved):
Works as expected

PR #582, opened 2024-07-10
0.32.3 hotfix

Starystars67:
This pr fixes BeamMP against BeamNG.drive 0.32.3 update.

@vulcan-dev please could you take a look at the Imgui chat that you did when it is convenient and raise a PR to resolve that down the road. :)


Olrosse (approved):
works as expected, arrow up chat history is disabled but chat now works

PR #583, opened 2024-07-10
ignition sync rewrite

Olrosse:
now syncs the ignition of all combustion engines individually, this fixes the B25 Mitchell engines not syncing

I've also added the spawnVehicleIgnitionLevel to the spawn packet so we can correctly sync the engine on resets, vehicles will still spawn running by default so we don't have cars driving around without the engine running

the engine and ignition state will now sync periodically so if the state is wrong when the car is spawned it will re-sync after a maximum of 10 seconds

i haven't synced the electric motors individually, However they seem to just automatically sync from just the ignition level sync, though i may add this later in case some mods control multiple motors individually

### DIFF
--- a/lua/ge/extensions/MPPowertrainGE.lua
+++ b/lua/ge/extensions/MPPowertrainGE.lua
@@ -50,12 +50,33 @@ local function applyLivePowertrain(data, serverVehicleID)
 end
 
 
+local function sendEngineData(data, gameVehicleID)
+	if MPGameNetwork.launcherConnected() then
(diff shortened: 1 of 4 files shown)


Starystars67:
Loose testing of this appears to work as expected now, though a little further testing should be done for any edge cases.
@WiserTixx maybe you would like to do some testing around this too?

Starystars67 (approved):

WiserTixx (approved):

PR #586, opened 2024-07-12
Fix imgui chat history

WiserTixx:
Fixed the imgui chat history for the new chat.


Starystars67 (approved):
Spot on, Does what it claims!

Olrosse (approved):
i do not understand ffi so i can't really review the code but it does appear to be working correctly

PR #592, opened 2024-07-17
Highlight queued players and selective queuing

### LINKED ISSUE #289: [Feature Request] Highlight player nametags who have a queued event
Highlight, in some way, players who currently have a queued event.

Maybe add a * to the user's nametag? (suggestion by @StanleyDudek) this should be discussed

**Additional context**
https://discord.com/channels/601558901657305098/777957916677242939/1054830270391533660

WiserTixx:
## The option to highlight players in the playerlist who have queued an event

Checkbox option
Option to enable highlighting queued players, enabled by default.  
<br>
<br>

image

Players that have a queue will light up in the playerlist.

Addresses issue #289.

---

## Selective queuing of players

https://github.com/user-attachments/assets/ece9cfc6-21e7-4a2b-b274-0dbe84e44157

Click on the name of the player whose queue you want to load.
Previously clicking on the name of a player put your camera at their vehicle, this functionality has been moved to right clicking on their name. This change can easily be reverted but in my opinion this is more intuitive.

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -897,12 +897,25 @@ end
 
 local function getQueueCounts()
 	local spawns, edits = 0, 0
+	local queuedPlayers = {}
+
+	local highlightQueuedPlayers = settings.getValue("highlightQueuedPlayers")
 
 	for serverVehicleID, vehicle in pairs(vehicles) do
-		if vehicle.spawnQueue then spawns = spawns + 1 end
-		if vehicle.editQueue then edits = edits + 1 end
+		if vehicle.spawnQueue then 
+			spawns = spawns + 1 
+			if highlightQueuedPlayers then
+				table.insert(queuedPlayers, vehicle.ownerID)
+			end
+		end
+		if vehicle.editQueue then 
+			edits = edits + 1 
+			if highlightQueuedPlayers then
+				table.insert(queuedPlayers, vehicle.ownerID)
+			end
+		end
 	end
-	return spawns, edits
+	return spawns, edits, queuedPlayers
 end
 
 
@@ -1868,6 +1881,26 @@ local function applyQueuedEvents()
 	--if currentVeh then be:enterVehicle(0, currentVeh) print("entered "..currentVeh:getJBeamFilename()) end -- Camera fix
 end
 
+local function applyPlayerQueues(playerID)
+	UI.updateQueue(getQueueCounts())
+	
+	for serverVehicleID, vehicle in pairs(vehicles) do
+		if vehicle.ownerID == playerID then
+			if vehicle.spawnQueue then
+				local data = vehicle.spawnQueue
+				vehicle.spawnQueue = nil
+				applyVehSpawn(data)
+			end
+			if vehicle.editQueue then
+				local data = vehicle.editQueue
+				vehicle.editQueue = nil
+				applyVehEdit(serverVehicleID, data)
+			end 
+		end
+	end
+
+	UI.updateQueue(getQueueCounts())
+end
 
 local function onUpdate(dt)
 	if MPGameNetwork and MPGameNetwork.launcherConnected() then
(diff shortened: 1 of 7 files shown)


StanleyDudek:
First of all thank you for putting in this work.

This looks very good.

I like these usability changes.

Question.

There exists a keybind for BeamMP to process the queue. Currently you can set the queue to not automatically process, and then use this keybind to process the queue whenever you want. It includes all queued spawns and edits from all players.

Do you think you could expand on what you have here to support a method of selecting which players you'd like to have update when you use the keybind to process the queue? This could also look like selecting players to *exclude* from the queue.

20dka:
Thank you, this is a nice feature

lua/ge/extensions/UI.lua
```diff
@@ -153,10 +153,24 @@ end
...
+        for key, player in pairs(MPVehicleGE.getPlayers()) do
+            if tableContains(queuedPlayers, key) then
+                guihooks.trigger("setQueueState", {playerID = key, state = true})
+            else
+                guihooks.trigger("setQueueState", {playerID = key, state = false})
+            end
+        end
+    else 
+        guihooks.trigger("resetQueueState")
+    end
```
20dka: please refactor this code to use the existing `setQueue` trigger, it is fine to update the entire queue state when an event gets applied, as other data probably changes too
```suggestion
	UIqueue = {spawnCount = spawnCount, editCount = editCount, queuedPlayers = (queuedPlayers or {})}
	UIqueue.show = spawnCount+editCount > 0
	sendQueue()
```

ui/modules/apps/BeamMP-PlayerList/app.js
```diff
@@ -174,7 +181,26 @@ app.controller("PlayerList", ['$scope', function ($scope) {
...
+		}
+	})
+
+	$scope.$on('resetQueueState', function(event, data) {
+		$scope.queuedPlayers = []
+		var rows = document.querySelectorAll('[id^="playerlist-row-"]');
+		for (let i = 0; i < rows.length; i++) {
+			rows[i].style.setProperty('background-color', 'transparent');
+		}
+	})
```
20dka: you can use the existing `setQueue` event and loop through the received data here, as opposed to sending an individual event for each queue status change

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -897,12 +897,25 @@ end
...
+			if highlightQueuedPlayers then
+				table.insert(queuedPlayers, vehicle.ownerID)
+			end
+		end
+		if vehicle.editQueue then 
+			edits = edits + 1 
+			if highlightQueuedPlayers then
+				table.insert(queuedPlayers, vehicle.ownerID)
+			end
+		end
```
20dka: this code can generate duplicate IDs, if one player has multiple queued events. you can avoid this by setting the table's keys temporarily such as:
```lua
if highlightQueuedPlayers then
	queuedPlayerKeys[vehicle.ownerID] = true
end
```
you will then probably want to iterate through the table and turn it in to an array of values to make the following code cleaner
```lua
local queuedPlayers = {}
for playerID in pairs(queuedPlayerKeys) do
	table.insert(queuedPlayers, playerID)
end
```

20dka (commented):
I left some comments and suggestions, please take a look and let me know if you agree with them

WiserTixx:
@20dka I've implemented your suggestions, could you please have another look?

WiserTixx:
@StanleyDudek I've gone ahead and written most of the code to implement your idea (I'm still working out how it should be visualized), however I'm not certain if this should be added with this PR.

20dka:
Thank you, i will be able to thoroughly test it tomorrow

Olrosse (approved):
tested and works as expected

Starystars67:
Tried and Tested. Works as expected.

Starystars67 (approved):

PR #602, opened 2024-08-25
fix for fatal lua error when joining servers

Olrosse:
this line was commented out because it caused mission markers to get stuck on the map when changing maps, this seems to no longer be happening, though please test just in case I've just been very lucky, adding the line back gets rid of the fatal lua error when joining servers


Starystars67 (approved):
Spot on, tested with Julian and all works as expected!
Thank you

PR #603, opened 2024-08-26
Position and velocity improvements

Olrosse:
Implemented a hybrid between cluster velocity and applyForceVector to get the performance benefit of cluster Velocity but without doors flying away when support beams are still attached,
best case scenario the cluster velocity can be up to 9400% or 94 times faster
This gave me roughly a 12% fps boost, tested on gridmap_v2 with 16 D series with angular velocity forced on.

Improved teleports by using cluster teleport and instant velocity in GE which doesn't damage vehicles and also doesn't trigger crash detection,
i also added one frame of prediction so it teleports to where it should be after the VE > GE queue delay.

Fixed vehicles sometimes flickering back to previous reset point on resets.

Reduced minimum force values and made velocity always apply if vehicle is traveling faster than 1 meter per second, this makes cars wonder less from their actual position and stops oversteery cars from crabbing

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -1553,10 +1553,11 @@ local function onServerVehicleResetted(serverVehicleID, data)
 			local veh = be:getObjectByID(gameVehicleID) -- Get associated vehicle
 			if veh then
 				local pr = jsonDecode(data) -- Decoded data
-				veh:reset()
 				if pr then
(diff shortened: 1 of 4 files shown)


stefan750:
Awesome work!

Code looks good, is there anything still open that needs to be done or is the pr ready?

Olrosse:
i don't have anything currently so it should be ready to merge :)

Mack29446 (approved):
lgtm

Starystars67 (approved):

mattnovelli:
Some great changes here, thanks!

PR #610, opened 2024-08-28
Changed coupler attach method

Olrosse:
changed coupling to attach the target node using obj:attachCoupler() rather than using activateAutoCoupling,
This stops the coupler debug blobs from activating on your screen when other are using trailers,
I've had to replace the activateAutoCoupling function to stop the blobs from activating when a remote vehicle with a trailer is reset.

This also helps the T series trailers to attach properly since it now tries to push and attach the coupler as long as it is within 10 meters


Starystars67:
Testing this with the Julian and indeed this does seem to work better now!

Starystars67 (approved):

WiserTixx (approved):

PR #615, opened 2024-09-12
remove * from custom role

NicholasMcMillan03:
You can literally upload malware to people who join your server, I think we can remove this star for the sake of people having cleaner roles, I could make a work around to make myself look like beamMP staff easily enough if I wanted to anyways


lionkor:
I agree the star is confusing

Starystars67 (approved):

PR #643, opened 2024-09-28
Check if the packet has a function

WiserTixx:
Avoids the big scary fatal lua errors that show when the serverlist packet gets improperly sent by the launcher.


lionkor:
My instinct is to not approve this change, because the real solution is fixing the issue.

Right now, this shows us when there is a problem. I guarantee (!!!) you that, if we turn off error toasts like this one, we will simply never fix the issue.

PR #648, opened 2024-10-09
Add functionality to check that the server cversion is actually compatible with the Launcher Version.

Starystars67:
This still needs to be fully tested with Launcher v2.2.X and Servers 3.6.X and below.


Starystars67:
@lionkor ready for when you are with the launcher and server <3

WiserTixx (approved):

PR #649, opened 2024-10-10
Instant paint sync + traffic paint sync

StanleyDudek:
The paint color pickers sync out immediately to all clients, and any traffic vehicles sync their new color when they teleport and reset.


WiserTixx:
See #653

PR #650, opened 2024-10-10
Add option for what to do when clicking on a player's name

WiserTixx:
This PR adds an option for which mouse button to use for applying a player's queued events in the playerlist, Allowing users to switch back to the old behaviour where if you clicked on a player's name your camera would switch to them.

image


Olrosse (approved):
tested, works as expected

PR #683, opened 2024-12-20
little changed UI

csv4211:
Fork updated

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -22,7 +22,7 @@ local launcherConnected = false
 local isConnecting = false
 local proxyPort = ""
 local launcherVersion = "" -- used only for the server list
-local modVersion = "4.13.11" -- the mod version
+local modVersion = "4.13.12" -- the mod version
 -- server
 
 local serverList -- server list JSON


csv4211:
sorry accidentally closed the last PR

ui/modModules/multiplayer/multiplayer.css
```diff
@@ -53,6 +53,57 @@ table#serversTable {
...
 }
 
+
+a.md-button.md-primary.md-raised, .md-button.md-primary.md-raised {
+  background-color: rgba(27,27,27,0.3);
+}
+
+a.md-button.md-primary.md-raised:not([disabled]):hover, .md-button.md-primary.md-raised:not([disabled]):hover { 
+  background-color: rgba(27,27,27,0.4);
+}
```
WiserTixx: I'd prefer to not have such generic styles, this might have impact on a game element somewhere which I'd like to avoid.

ui/modModules/multiplayer/multiplayer.css
```diff
@@ -53,6 +53,57 @@ table#serversTable {
...
+
+.login-button-block {
+  display:flex;
+  align-content: space-evenly;
+  width: 400px; 
+  width: auto;
```
WiserTixx: Are there meant to be 2 width values here?

WiserTixx (changes requested):

ui/modModules/multiplayer/multiplayer.css
csv4211: fix it now, line 68 shouldn't have been there

ui/modModules/multiplayer/multiplayer.css
csv4211: I made a mistake, there should be classes ".md-button.md-primary.md-raised" and ".md-button.md-primary.md-raised:not([disabled]):hover"

csv4211: I tried to get rid of such common styles, but the direct button goes as an <a> element and not as an <md-button>, which is why I have to use the style "a.md-button.md-primary.md-raised: not([disabled]):hover" instead of ".md-button.md-primary.md-raised:not([disabled]):hover" - which matches the rest of the buttons,
I tried to add a separate class for the direct connection button, but its background still did not change when activated

WiserTixx: Okay, and why did you make this change? Testing the buttons myself they seem to all act the same way.

csv4211: image
in the first screenshot these styles are indicated 
`
a.md-button.md-primary.md-raised, .md-button.md-primary.md-raised {
  background-color: rgba(27,27,27,0.3);
}

a.md-button.md-primary.md-raised:not([disabled]):hover, .md-button.md-primary.md-raised:not([disabled]):hover { 
  background-color: rgba(27,27,27,0.4);
}
`

image
and on the second one such 
`
.md-button.md-primary.md-raised {
  background-color: rgba(27,27,27,0.3);
}

.md-button.md-primary.md-raised:not([disabled]):hover { 
  background-color: rgba(27,27,27,0.4);
}
`

maybe it's just me who has some bugs, but it should be like in the first screenshot

WiserTixx: Why should it be like the first one?

WiserTixx:
It looks like you've changed a few colours, are those all stylistic changes?

PR #704, opened 2025-02-27
Playerlist context menu

WiserTixx:
# Playerlist context menu

Adds a context menu to the playerlist when right clicking a username.

image

## Custom buttons
Custom buttons can be added to the context menu with simple lua code.
```lua
UI.getCustomPlayerlistButtons()['Cool button'] = function(name, id) print(name) print(id) end
```

image

## Settings
This PR replaces the `queueWithLMB` setting with `playerlistLeftclick`, to allow for more assignable actions. Code that migrates the old setting is included.

image


Olrosse (approved):
Tested, works like expected, i like that the click function is customizable

PR #714, opened 2025-03-30
Add "patching" behaviour to AddEventHandler, add RemoveEventHandler and pcall called event functions

OfficialLambdax:
This solves the following issue.

```lua
local function myEventFunction()
   -- do something
end

AddEventHandler("myEvent", myEventFunction)
AddEventHandler("myEvent", myEventFunction)
```

This function will now be twice in `eventTriggers`. This issue is easy to trigger when the extension that listens to beammp events is hotreloaded.

This behaviour is changed with this PR. AddEventHandler will now lookup if it already knows of this event and if so patch the function instead of just adding it again. Multiple functions can still listen to the same event. The check is performed via `debug.getinfo(2).source`

So this introduces the limitation that a lua file can only have 1 function listen per event it wants to listen to
```lua
local function myEventFunction()
   -- do something
end

local function myEventFunction2()
   -- do something
end

AddEventHandler("myEvent", myEventFunction)
AddEventHandler("myEvent", myEventFunction) -- will replace the previous
AddEventHandler("myEvent2", myEventFunction2) -- will work
AddEventHandler("myEvent", myEventFunction2) -- will also work
```

grafik

### DIFF
--- a/lua/ge/extensions/MPGameNetwork.lua
+++ b/lua/ge/extensions/MPGameNetwork.lua
@@ -189,13 +189,14 @@ function TriggerClientEvent(name, data)
 end
 
 --- Adds an event handler for the specified event name and function.
--- @tparam string name - The name of the event
+-- @tparam string event_name - The name of the event
 -- @tparam function func - The event handler function
+-- @tparam string name - The internal name (optional)
 -- @usage AddEventHandler(`<name>`, `<function>`)
 -- @usage if AddEventHandler then AddEventHandler(`<name>`, `<function>`) end -- if your mod is also singleplayer available
-function AddEventHandler(name, func)
-	if name == nil or type(name) ~= "string" or (type(name) == "string" and name:len() == 0) then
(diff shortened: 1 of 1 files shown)


OfficialLambdax:
Further things to be considered
- [x] Add `RemoveEventHandler(event_name)`?
- [ ] and/or `MPGameNetwork.removeEventHandler(event_name)`?
- [x] Add optional arg to `AddEventHandler(event_name, func, >>name<<)` to perform `local source = name or debug.getinfo(2).source` which would make the previous `RemoveEventHandler` callable from everywhere?
```lua
local function myEventFunction()
  -- do something
end

AddEventHandler("myEvent", myEventFunction, "funny random name")
RemoveEventHandler("funny random name")
```

20dka:
I vote for adding a Remove function, keep it a global for consistency.
The optional arg is also a good idea

WiserTixx (approved):

PR #724, opened 2025-04-06
Add option to disable the mod if the launcher isn't open

WiserTixx:
If the game was started without the launcher and the intention is to play singleplayer, this option can be used. And this will not be an inconvenience if the user does want to play BeamMP as the launcher enables the mod before launching the game.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -92,7 +92,11 @@ local function send(s)
 			authResult = {}
 			guihooks.trigger("authReceived", authResult)
 		elseif error == "Socket is not connected" then
-
+			if settings.getValue("disableWithoutLauncher") then
+				core_modmanager.deactivateMod("beammp")
+				core_modmanager.deactivateMod("multiplayerbeammp")
+				log('W', 'send', 'Launcher not connected! Deactivating mod!')
+			end
 		else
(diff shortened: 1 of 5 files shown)


lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -92,7 +92,11 @@ local function send(s)
 			authResult = {}
 			guihooks.trigger("authReceived", authResult)
 		elseif error == "Socket is not connected" then
-
+			if settings.getValue("disableWithoutLauncher") then
+				core_modmanager.deactivateMod("beammp")
+				core_modmanager.deactivateMod("multiplayerbeammp")
```
OfficialLambdax: This is called repeatably every 1.5 seconds for up to 11 times

OfficialLambdax (approved):
Ok to me, thank you @WiserTixx !
This PR produces a potentialy unwanted side effect. Should this option be enabled it is not (without file edits) possible to get beammp loaded in singleplayer. because it essentially directly disables itself again.

I leave the merge to you

stefan750:
I agree with @OfficialLambdax here that might be an unintended side effect, maybe a solution would be to not disable it immediately, but only disable it when loading into a map without the launcher connected.

Starystars67:
I think I agree here or that it should be decided on some other additional factor but I am not sure what just yet.

PR #725, opened 2025-04-07
Update MPModManager.lua to prevent mods being disabled

Ausdboss:
I deleted line 103. It prevents mods being disabled when launching BeamNG through the BeamMP launcher and joining a server. Many people have asked on Reddit for the past couple years asking if there was a way to stop disabling of mods so they don't have to enable UI mods again, or so friends can play in Vanilla servers with all the mods they use and it syncs perfectly, as well as adding mods to servers that don't have all the mods you want. In my case it's also to save space so you don't have to re-download mods you already have for a server and so I don't have to enable a script in the editor and then re activate my 82 mods with my friend every single time we play, takes like 20 mins, and prevented us from actually playing the game. Now we have been playing non stop and love it! I want to share the love with others! I think this would work best as an on and off switch so you can keep the current functionality or use this new method of mods not disabling. Has been a GAME CHANGER for me and have had no issues! Hope you guys enjoy!

Screenshot 2025-04-06 193823

### DIFF
--- a/lua/ge/extensions/MPModManager.lua
+++ b/lua/ge/extensions/MPModManager.lua
@@ -100,7 +100,7 @@ local function checkMod(mod) --TODO: might have a flaw with repo mods as their n
 
 	if not modAllowed and mod.active then
 		log('W', 'checkMod', "This mod should not be running: "..modname)
-		core_modmanager.deactivateMod(modname)
+	-- Deleted Line to prevent disabling of mods -- Would prefer a switch to turn on and off -- Ausdboss
 		if mod.dirname == '/mods/multiplayer/' then
 			core_modmanager.deleteMod(modname)
 		end


WiserTixx:
This is intentional behaviour, only the server's mods are activated to prevent desync and to increase the session quality for everyone on the server, as they'll be able to see what is going on instead of 2 black blobs (missing jbeam) flying around.
You can add mods to the launchers cache usually located at `%appdata%/BeamMP-Launcher/Resources` or in the directory you open the launcher from. The launcher will check if it already has the mod before trying to download it again.

My advise is to [create your own server](https://docs.beammp.com/server/create-a-server/) and add the mods you want to play with, this will ensure that your friend will always have the same mods and it will minimise desync.

Ausdboss:
That does make sense. I know BeamNG categorizes mods, Could there be a way to have an option in settings of BeamMP to allow UI and camera mods etc, to stay enabled and just disable car mods? The main thing I'd like to see is a way for BeamMP to re-enable the mods you had on when going back to single player. in my case 82/146 mods activated, always a bad time re-activating all of those, combing through each mod. I'll try and see if I can get something like that working. Maybe use the activate all mods script to categorize the mods you last had enabled and only re-enable those said mods.

WiserTixx:
> Could there be a way to have an option in settings of BeamMP to allow UI and camera mods etc, to stay enabled and just disable car mods? 

Yes, however that category (UI, Camera, ..) is purely visual. The mod creator can still iirc add vehicles and lua code in a UI mod.

> I'll try and see if I can get something like that working. Maybe use the activate all mods script to categorize the mods you last had enabled and only re-enable those said mods.

That would be great, thank you. You might also be able to store the mods that have to be deactivated before connecting to a server in an array, which you can then loop over after exiting the session, and reactivate them.
Here's a [PR that adds a setting](https://github.com/BeamMP/BeamMP/pull/724) as an example.

OfficialLambdax:
Yeah there are a multitude of factors that made us decide to auto disable mods. One of them being that the community owners that fuel the community with their servers asked us to have this. And as much as i like to allow certain mods, by eg categorizing them, that is unfortunately alot easier said then done.

@WiserTixx about the mods reinstating, we could totally cache all active mods before we disable them (on server join) and then reinstate that after leave

But in any case, we are not intending to revert the mod disabling feature at this moment.

PR #733, opened 2025-04-25
Fix fatal in MPCoreNetwork.onUpdate() from processing invalid data

OfficialLambdax:
The game own socket.tcp() class seems to be configured to auto split received data by newlines which are then given back from the `:receive()` method used by beammp. Unfortunately the class fails todo so at times and either doesnt split data at all or merges it with other.

This leads to corrupted and unfixable data on the receiving end. i cannot find the origin of this behaviour and would change it if i could, but given that the Launcher -> Game protocol works with newlines as packet seperators i see no way to cleanly seperate packets any other way.

As such this is merely a troubleshoot. Please be aware that this fix throws the received data away (not that it would help if it didnt)

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -629,17 +629,19 @@ local function onUpdate(dt)
 		if TCPLauncherSocket ~= nop then
 			while(true) do
 				local received, stat, partial = TCPLauncherSocket:receive()
-				if not received or received == "" then
+				if not received or received:len() == 0 then
 					break
-				end
-				if settings.getValue("showDebugOutput") then -- TODO: add option to filter out heartbeat packets
-					log('M', 'onUpdate', 'Receiving Data ('..#received..'): '..received)
 				end
 
 				-- break it up into code + data
 				local code = string.sub(received, 1, 1)
 				local data = string.sub(received, 2)
-				HandleNetwork[code](data)
+				
+				if settings.getValue("showDebugOutput") and code ~= 'A' then
+					log('M', 'onUpdate', 'Receiving Data ([' .. code .. ']' .. #received .. '): ' .. received)
+				end
+				
+				if HandleNetwork[code] then HandleNetwork[code](data) end
 			end
 		end
 		--================================ SECONDS TIMER ================================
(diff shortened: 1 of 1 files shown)


OfficialLambdax:
This essentially gets rid of the big error screen that players often can come accross when they open the serverlist

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -629,17 +629,23 @@ local function onUpdate(dt)
...
-				if settings.getValue("showDebugOutput") then -- TODO: add option to filter out heartbeat packets
-					log('M', 'onUpdate', 'Receiving Data ('..#received..'): '..received)
-				end
 
 				-- break it up into code + data
 				local code = string.sub(received, 1, 1)
 				local data = string.sub(received, 2)
-				HandleNetwork[code](data)
+				
+				if settings.getValue("showDebugOutput") and code ~= 'A' then
```
WiserTixx: Could there be a settings for ignoring heartbeat packets? I feel like it can be useful to see them in some cases. This change can also be revert and then added back in a different pr for packet log filtering with say filtering position packets and electrics as well.

lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -629,17 +629,23 @@ local function onUpdate(dt)
...
+					log('M', 'onUpdate', 'Receiving Data ([' .. code .. ']' .. #received .. '): ' .. received)
+				end
+				
+				if not HandleNetwork[code] then
+					log('E', 'onUpdate', 'Received corrupted packet fragment. Ignoring data.')
```
WiserTixx: There was already a PR for this https://github.com/BeamMP/BeamMP/pull/643, but the general consensus was we'd rather actually fix the issue. 
It would also be nice to still log the data here, if I'm changing or adding things to the launcher <-> game protocol I'd like to know whether I've messed something up or not.

WiserTixx (changes requested):

lua/ge/extensions/MPCoreNetwork.lua
OfficialLambdax: Good point im reverting

lua/ge/extensions/MPCoreNetwork.lua
OfficialLambdax: Good point and yes fixing it would be alot better, but i cant do that with the current protocol design on the client side only. Imma add the packet fragment to the log

OfficialLambdax:
grafik

WiserTixx (approved):

PR #754, opened 2025-06-17
Fix selected row in serverlist having a bonus column because of .focus-visible class since 0.36

boubouleuh:
This fix the weird behavior of the selected row in the serverlist since 0.36, i dont really know why this beamng class apply on the row like this though


boubouleuh:
seems to be an issue that only me have for some reasons, mb 🫡

boubouleuh:
Reopening this since it happens to other people as well

PR #761, opened 2025-07-15
Fix Launcher -> Game message loss (v2 Networking)

### LINKED ISSUE #763: Native Linux Client - Inconsistent Event Queuing.
I run a BeamMP server for myself, my kids, and my friends. My household is a mix of mostly Linux with a couple of Windows boxes. My friends are all Windows users. Until fairly recently I had the Linux players running the Windows BeamMP launcher via Steam Tinker Launcher, which had worked fine until recent updates around December apparently broke it. Last week I pulled the source, and built it. No issues building it, and no issues running it. Everything appears to work. We run the random config and the outbreak mod on my server, which means lots of event queuing every round. This is where we discovered an issue.

Roughly 3 out of 4 times a car is loaded from a different client, no event will queue on the native Linux clients. It's not all of them, it might hit one while not another. It might successfully queue 2 or 3 times in a row, then not queue anything 3 or 4 times in a row. Initially we thought it might be related to the random config mod, and stopped using that; manually picking the cars from the vehicle menu. The issue remains, and events fails to queue semi randomly, but only for native Linux clients; Windows clients do not fail at all.

When the event fails to queue it may still track the players car for the native Linux clients, or it may fully de-sync, and now that players car appears to be stationary, but is in fact moving for every other client connected to the server. Having that player load cars until the event queues for the native Linux client restores the sync. While that works as a work around, having more than 1 native Linux client connected means that it may fix it for one client and break the other.

No errors are output or logged on the server or the clients. I can reliably reproduce the issue across multiple physical computers, running a variety of Linux distros. My primary one is Arch on kernel 6.11.11 running on a Ryzen 7 5800X, 64GB of ram, and a RX 6900 XT GPU.

If there is some debug logging I can enable at the server and client, I would be happy to do so and provide all the logs.

### LINKED ISSUE #764: Serverlist Parsing issue
The LUA does not know when serverlist entries are sent split up, causing Fatal LUA Errors to appear. This can happen when the Server has too many characters in the entry body.

Barinzaya:
Since [the launcher sends the message data and the ending newline via separate calls to send](https://github.com/BeamMP/BeamMP-Launcher/blob/e7cfb6e406a96d2f57ab5c35a69439bf447da63a/src/Network/GlobalHandler.cpp#L83), it's possible for the game to receive a message's data but not the following newline in a call to `receive`. This is considered a `timeout` error and the partial message data is returned in the `partial` result--but this was being ignored, resulting in the message being effectively lost. It is already read, and thus isn't included in future reads.

This seems to be a much more prevalent issue when running the client natively on Linux, and seems to be more noticeable in `MPGameNetwork`, resulting in vehicle events, chat messages, server-triggered events, and possibly more being silently dropped.

This should address https://github.com/BeamMP/BeamMP/issues/763.

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -21,6 +21,7 @@ local ltn12 = require("ltn12")
 local launcherConnected = false
 local isConnecting = false
 local proxyPort = ""
+local socketPartialData
 local launcherVersion = "" -- used only for the server list
 local modVersion = "4.16.0" -- the mod version
 -- server
@@ -120,6 +121,7 @@ local function connectToLauncher(silent)
 	isConnecting = true
 	if not silent then log('W', 'connectToLauncher', "connectToLauncher called! Current connection status: "..tostring(launcherConnected)) end
 	if not launcherConnected and not mp_core then
+		socketPartialData = nil
 		TCPLauncherSocket = socket.tcp()
 		TCPLauncherSocket:setoption("keepalive", true) -- Keepalive to avoid connection closing too quickly
 		TCPLauncherSocket:settimeout(0) -- Set timeout to 0 to avoid freezing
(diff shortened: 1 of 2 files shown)


WiserTixx:
Also possible fix for https://github.com/BeamMP/BeamMP/issues/764

WiserTixx (approved):

PR #770, opened 2025-08-02
Make kick dialog text and title white

### LINKED ISSUE #769: [Bug] Kick/Disconnect Message - Text Color
Black text in a gray box when you're disconnected is difficult to read.

Image

We should make the text color white.

If you were disconnected by a moderator, consider making the text red and maybe bold.

WiserTixx:
This PR makes the kick dialog text white on grey instead of black on dark grey.
image


Olrosse (approved):
tested, works as expected

PR #787, opened 2025-09-17
[Fix] Change from debugDrawer:drawTextAdvanced to ffi.C.BNG_DBG_DRAW_TextAdvanced to regain control over Z rendering + option

### LINKED ISSUE #777: [Bug] [0.37/4.18] Name Tags hide behind scene + wrong alpha
In-Game name tags are now bugged in that they are lacking the alpha, they are also rendering behind items like distant scenery, trees and barriers.

Olrosse:
Nametags now render on top of everything again by default, but i added an option to enable the z buffer again

fixed render alpha by switching to the FFI drawer, the other one seems bugged? the FFI drawer should also be a lot faster and produce a lot less less garbage

the nametags seem to be a bit more transparent now, but I'm not sure if that's just me, maybe something to double check and adjust if necessary

### DIFF
--- a/lua/ge/extensions/MPVehicleGE.lua
+++ b/lua/ge/extensions/MPVehicleGE.lua
@@ -32,6 +32,10 @@ local original_spawnNewVehicle
 local original_replaceVehicle
 local original_spawnDefault
 
+-- the debugDrawer functions without draw in the name calls FFI directly, these runs a lot faster and produces no garbage,
+-- however it requires xyz to be input independently, check lua\common\utils\debugDraw.lua to see the available functions
+local debugDrawer2 = require("utils/debugDraw")
+local drawTextAdvanced = debugDrawer2.TextAdvanced
 
 --- Contains Information about Backend authorized Roles
 -- @table roleToInfo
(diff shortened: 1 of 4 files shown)


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -32,6 +32,10 @@ local original_spawnNewVehicle
 local original_replaceVehicle
 local original_spawnDefault
 
+-- the debugDrawer functions without draw in the name calls FFI directly, these runs a lot faster and produces no garbage,
+-- however it requires xyz to be input independently, check lua\common\utils\debugDraw.lua to see the available functions
+local debugDrawer2 = require("utils/debugDraw")
+local drawTextAdvanced = debugDrawer2.TextAdvanced
```
OfficialLambdax: Could instead directly call ffi.C.BNG_DBG_DRAW_TextAdvanced, because debugDrawer2.TextAdvanced performs a full param check that i think is uneccesary

OfficialLambdax (changes requested):
Besides the one mentioned change that i suggest, the pr works nicely.

I tested
- the normal nametag appearance
- the special role nametag appearance
- the new option this pr introduces
- the nametag fading options that are already inbuilt
- that spectators are properly displayed

This pr is good to go

Starystars67:
What do we think about merging the new functionality to be part of the fade nametags setting as well so that they better hide at distance and then this PR which resolves the change as part of 0.37 would become a show nametags globally?

OfficialLambdax:
Not sure what you mean. Do you mean to put it into the nametag fading settings group? so that this new option only shows up when nametag fading is enabled?

Starystars67:
So what I mean is, lets try and build out two options for nametags:
1. Name tags always on (uses the solution in this PR, mirroring how nametags have been normally).
2. Name tags hide behind items in the map and automatically fade out over distance.

Thoughts?

OfficialLambdax:
Both functionalities are present in the mod.
1. Is archived with this PR
2. Is nonetheless a feature of BeamMP with the namtag fading option thats already present

Am i missing something?

OfficialLambdax:
Awesome work @Olrosse, this pr is good to go, just waiting for @Starystars67 to clarify

OfficialLambdax (approved):
Ready for merge

PR #789, opened 2025-09-18
General performance improvements

Starystars67:
solution to performance issues on west coast thanks to feedback from the team and community.

Also some minor improvements to performance ~1 or 2 fps at best. Will continue to chip away at it.


lua/ge/extensions/MPCoreNetwork.lua
```diff
@@ -59,6 +59,9 @@ local heartbeatTimer = 0
 local reconnectTimer = 0
 local reconnectAttempt = 0
 
+-- defined vars to prevent lookups in main loop
+local setting_SHOWDEBUGOUTPUT = settings.getValue("showDebugOutput")
```
WiserTixx, line 63: Doesn't this force the user to reload lua/restart their game when they update this setting?
I often toggle it during a session, so having to reconnect every time would be quite bothering.

WiserTixx (changes requested):

lua/ge/extensions/MPCoreNetwork.lua
Starystars67: It would however given how "spammy" it is and that it is only likely to be used during debug activity I thought it might be okay. 

Alternatively lets see if we can add in a watch for setting change and then update it from that rather than checking it every frame.

Olrosse: there is an "onSettingsChanged" hook, if that works how I think i does it could solve both of those

Starystars67: > there is an "onSettingsChanged" hook, if that works how I think i does it could solve both of those

That would be perfect if that's the case!

OfficialLambdax: settings.getValue() is cheap.

```lua
local function getValue(key, defaultValue)
  local value = getValues()[key]
  if value == nil then
    return defaultValue
  end
  return value
end
```

Starystars67: > settings.getValue() is cheap.
> 
> ```lua
> local function getValue(key, defaultValue)
>   local value = getValues()[key]
>   if value == nil then
>     return defaultValue
>   end
>   return value
> end
> ```

It is but do we need to waste any amount of performance just for the sake of it?

Starystars67: This should be resolved now with the latest commit as I added the onSettingsChanged hook proposed by Olrosse.

OfficialLambdax: A single setting pull per frame is unlikely to even be measurable. The mod is wasting performance in other areas. eg this is a performance killer

Two wasted vec objects in both cases. The vec3 object is instantiated, initialized and then dropped. dropping alot of objects causes stutter.
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2255
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2291



Heres an object instantiated, used once or twice and then dropped again. Not gonna list all these cases, because we do this everywhere.
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2472
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2440
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2377



And heres a big one. targetPosition and node1 is initiated during runtime somewhere and then we do an operation with it. That operation generates another object.
https://github.com/BeamMP/BeamMP/blob/05026b61b38758347f75c54cf9e54dc771113856/lua/ge/extensions/MPVehicleGE.lua#L2123



To demonstrate what i mean
```lua
local function dirBetweenTwoVehicles(veh_1, veh2)
  local pos_1 = veh_1:getPosition() -- vec3 instantiated
  local pos_2 = veh_2:getPosition() -- vec3 instantiated

  local div = pos_1 - pos_2 -- vec3 instantiated
  local normalized = div:normalized() -- vec3 instantiated

  return normalized

  -- or in short
  return (veh_1:getPosition() - veh_2:getPosition()):normalized()
end
```
4 vec objects instantiated and wasted all within one simple math operation (Thats about 1kb of memory garbage btw.). this is an issue, not a single setting pull.

Starystars67: Great feedback @OfficialLambdax and we absolutely should be looking at this too!

I just happened to be looking into something else and came across the changes I made but if there are some of you who are okay with also working on this then I welcome it with open arms!

Are you likely to make these yourself or shall I work through these?

OfficialLambdax: This is what the game devs want us modders todo. In short, instantiate objs on load and then reuse them on runtime, while using methods from game engine objects that dont create new objecte (aka the XYZ alternatives for vec3 returns) and to not use any of the normal math operators (+ - * /) when they have the effect of creating a new obj https://documentation.beamng.com/modding/programming/performance/#avoid-garbage-collection

Ive played with it, by translating a present script into the whole obj reuse system that the game offers for vec3. its hardly readable and something that i for myself have not used again after this
https://github.com/OfficialLambdax/BeamNG-MapModules/blob/main/levels/your_map_name/vlua/portals.lua

The solution i found for writing zero garbage code was to make my own vector 3 implementation and to also when possible to reuse tables if they always have fixed sizes. To give an idea, this whole renderer generates zero garbage, even tho it has to know 2 positions per box, of which are ~10k in a chunk, of which 100 exist per region (aka ~2 million positions). From gen to render, zero garbage.
https://www.youtube.com/watch?v=LIcnMCPB5IE

Unfortunately i havent released a mod using this method yet, but i do have a private repo of a mod that can demonstrate on how to write zero garbage code.

Olrosse: yep we should do all those, i did similar optimizations to my drafting mod and additionally used ffi structs for all my wind positions, which can be over 100 at higher speeds, it improved the performance drastically, especially after a few minutes of run time where the old version ran into garbage issues, I'm not sure if we use enough vec3s for it to be a huge difference, but it will be a good improvement either way

PR #796, opened 2025-09-21
hide mission markers while in multiplayer session

Olrosse:
This is also an alternative fix for the Minimap lag without setting the state to freeroam,

at first i tried just patching that function like suggested, and while that worked for the lag issue, it also had the problem of the missions being visible, this workaround should fix both

The game seems to have several places where it sets them to visible,
setMarkersVisibleTemporary worked for markers when the map is not open,
however if you quick traveled to one it would still render that one, and you could start that challenge even after clearing the marker table,
Removing onPreRender and hiding them manually is the only way i found to solved that so far, while doing it this way feels a bit hacky, it should be fine for now.

The markers still render in the big map, so you can still use the travel points, however as soon as you exit the map or quick travel they all disappear again.

All of this reverts as you leave the session, so the markers and map should still work as expected outside of multiplayer

### DIFF
--- a/lua/ge/extensions/MPCoreNetwork.lua
+++ b/lua/ge/extensions/MPCoreNetwork.lua
@@ -707,7 +707,7 @@ runPostJoin = function() -- gets called once loaded into a map
 		MPGameNetwork.connectToLauncher()
 		log('W', 'runPostJoin', 'isGoingMpSession = false')
 		isGoingMpSession = false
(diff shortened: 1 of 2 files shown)


Starystars67:
Just based on what you have written here this certainly sounds like a great solution to the issue!

Nice work!

WiserTixx:
This issue was fixed in BeamNG 0.37.5 and #804

PR #801, opened 2025-09-23
Revised the loading screen before we begin loading the game :)

Starystars67:
image

looking for feedback ofc :)

### DIFF
--- a/ui/modModules/multiplayer/multiplayer.html
+++ b/ui/modModules/multiplayer/multiplayer.html
@@ -6,20 +6,20 @@
   <div id="LoadingBody" style="">
     <md-button style="right:0; top:0; margin:0;" class="button close-button" ng-class="" ng-click="multiplayer.closeLoadingPopup()"> X </md-button>
     <h1>{{:: 'ui.multiplayer.connectingToServer' | translate}}</h1>
-    <hr ng-if="multiplayer.loadingStatus != '' || multiplayer.downloadingMods.length > 0"/>
+    <hr id="loadingstatus-divider"/>
     <span id="OriginalLoadingStatus" hidden style="height: 0px;"></span>
     <span id="LoadingStatus" hidden style="height: 0px;">{{multiplayer.loadingStatus}}</span>
     <div id="mod-download-list" ng-if="multiplayer.downloadingMods.length > 0">
       <div ng-repeat="mod in multiplayer.downloadingMods" class="mod-download-row" style="align-items: center; margin-bottom: 10px;">
         <span style="flex: 1;">
-          <code>{{mod.name}}</code>
+          <code>{{mod.number}} - {{mod.name}}</code>
           {{ ' - ' }}
-          <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
+          <span style="color: darkgray;font-size: 0.8em;">{{$index === 0 ? mod.speed : 'Downloaded'}}</span>
         </span>
-        <div style="flex: 2;" ng-if="mod.progress != 100">
+        <div style="flex: 2;" ng-if="$index === 0 && mod.progress != 100">
           <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
         </div>
-        <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
+        <span ng-if="$index != 0" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
           <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
         </span>
       </div>
--- a/ui/modModules/multiplayer/multiplayer.html
+++ b/ui/modModules/multiplayer/multiplayer.html
@@ -14,7 +14,7 @@
         <span style="flex: 1;">
           <code>{{mod.number}} - {{mod.name}}</code>
           {{ ' - ' }}
-          <span style="color: darkgray;font-size: 0.8em;">{{$index === 0 ? mod.speed : 'Downloaded'}}</span>
+          <span style="color: darkgray;font-size: 0.8em;">{{$index === 0 ? mod.speed : ('ui.multiplayer.download.downloaded' | translate)}}</span>
         </span>
         <div style="flex: 2;" ng-if="$index === 0 && mod.progress != 100">
           <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
--- a/ui/modModules/multiplayer/multiplayer.html
+++ b/ui/modModules/multiplayer/multiplayer.html
@@ -11,17 +11,17 @@
     <span id="LoadingStatus" hidden style="height: 0px;"></span>
     <div id="mod-download-list">
       <div ng-repeat="mod in multiplayer.downloadingMods" class="mod-download-row" style="align-items: center; margin-bottom: 10px;">
-      <span style="flex: 1;">
-        <code>{{mod.number}}: {{mod.name}}</code>
-        {{ ' - ' }}
-        <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
-      </span>
-      <div style="flex: 2;" ng-if="mod.progress != 100">
-        <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent rounded-progress"></md-progress-linear>
-      </div>
-      <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
-        <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
-      </span>
+        <span style="flex: 1;">
+          <code>{{mod.name}}</code>
+          {{ ' - ' }}
+          <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
+        </span>
+        <div style="flex: 2;" ng-if="mod.progress != 100">
+          <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
+        </div>
+        <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
+          <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
+        </span>
       </div>
       <div style="
       pointer-events: none;
--- a/ui/modModules/multiplayer/multiplayer.html
+++ b/ui/modModules/multiplayer/multiplayer.html
@@ -12,7 +12,7 @@
     <div id="mod-download-list">
       <div ng-repeat="mod in multiplayer.downloadingMods" class="mod-download-row" style="align-items: center; margin-bottom: 10px;">
       <span style="flex: 1;">
-        <code>{{mod.name}}</code>
+        <code>{{mod.number}}: {{mod.name}}</code>
         {{ ' - ' }}
         <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
       </span>
--- a/ui/modModules/multiplayer/multiplayer.html
+++ b/ui/modModules/multiplayer/multiplayer.html
@@ -17,7 +17,7 @@
         <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
       </span>
       <div style="flex: 2;" ng-if="mod.progress != 100">
-        <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
+        <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent rounded-progress"></md-progress-linear>
       </div>
       <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
         <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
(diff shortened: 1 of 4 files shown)


TheHawk09:
That looks good! Adding a border-radius of 2px to the progress bar would be beneficial. Additionally, I believe the "X" button would look better with a larger border-radius.

WiserTixx:
> That looks good! Adding a border-radius of 2px to the progress bar would be beneficial. Additionally, I believe the "X" button would look better with a larger border-radius.

@Starystars67 What do you think?

Starystars67:
Would you be okay with sending an image of what you have in mind?

TheHawk09:
492915495-3c8f460f-e956-410b-aa8f-b2d9aaea3ab9
@Starystars67 Here is a rough idea of what I envisioned. The "X" button didn't turn out quite right, but that's okay

WiserTixx:
In my opinion the rounded progress bar looks better but the "X" should stay the same to match the rounding of the container.

Olrosse:
personally i think either bar looks fine, the rounded loading looks nice but i also think the square one is fine, However i agree with Tixx on the "X"

TheHawk09:
Yep, sounds good. Now that I look at it, I agree with keeping the "X" as-is.

Olrosse:
One thing i noticed is that this one is not showing how many mods you have left to download, that should probably be added back before merge

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+			});
 		} else {
-			document.getElementById('LoadingStatus').innerText = data.message;
+			if (document.getElementById('LoadingStatus').innerText != data.message) console.log(data.message)
+			if (data.message == "done") {
+				document.getElementById('LoadingStatus').innerText = "Done";
+				lastModInfo = ''
+			} else {
+				document.getElementById('LoadingStatus').innerText = data.message;
+			}
```
roitto, line 806: If you use the `text-transform: capitalize;` style for the LoadingStatus element, you could do something like this instead:


```suggestion
			if (data.message == "done") {
				lastModInfo = '';
			}

			document.getElementById('LoadingStatus').innerText = data.message;
```

.. Or is there some case where we don't want to show the data.message with capitalized text?

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+			if (matches) {
+				modNumber = matches[1];
+				modName = matches[2];
+			}
+			//console.log(`Mod ${modNumber}: ${modName} - Loading`);
+
+			// Update current downloading mod info and if complete then push this mod into the downloaded mods info
+			$scope.$apply(function() {
+				// Update or add the current mod being downloaded
+				let existingMod = vm.downloadingMods.find(mod => mod.name === modName);
```
roitto: No need to use `let`, since this variable is not reassigned.

```suggestion
				const existingMod = vm.downloadingMods.find(mod => mod.name === modName);
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+					existingMod.progress = progress;
+					existingMod.speed = speed;
+				} else {
+					// add this new mod to the beginning of the array
+					vm.downloadingMods = [{ number: modNumber, name: modName, progress: progress, speed: speed }, ...vm.downloadingMods];
+				}
+
+				// If we switched to a new mod, mark the last one as done
+				if (lastModInfo != '' && lastModInfo != modName) {
+					let lastMod = vm.downloadingMods.find(mod => mod.name === lastModInfo);
```
roitto: No need to use `let`, since this variable is not reassigned.

```suggestion
					const lastMod = vm.downloadingMods.find(mod => mod.name === lastModInfo);
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+				modName = matches[2];
+				progress = matches[3];
+				speed = matches[4] || '...';
+			}
+			//console.log(`Mod ${modNumber}: ${modName} - ${progress}% at ${speed}`);
+
+			// Update current downloading mod info and if complete then push this mod into the downloaded mods info
+			$scope.$apply(function() {
+				// Update or add the current mod being downloaded
+				let existingMod = vm.downloadingMods.find(mod => mod.name === modName);
```
roitto: No need to use `let`, since this variable is not reassigned.

```suggestion
				const existingMod = vm.downloadingMods.find(mod => mod.name === modName);
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+				if (existingMod) {
+					existingMod.progress = progress;
+					existingMod.speed = speed;
+				} else {
+					// add this new mod to the beginning of the array
+					vm.downloadingMods = [{ number: modNumber, name: modName, progress: progress, speed: speed }, ...vm.downloadingMods];
```
roitto, line 764: Dunno about angular handles reactivity, but would it be better to just do unshift so we don't need to create new array each time?


```suggestion
					vm.downloadingMods.unshift({ number: modNumber, name: modName, progress: progress, speed: speed });
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+
+			// Update current downloading mod info and if complete then push this mod into the downloaded mods info
+			$scope.$apply(function() {
+				// Update or add the current mod being downloaded
+				let existingMod = vm.downloadingMods.find(mod => mod.name === modName);
+				if (existingMod) {
+					existingMod.progress = '100';
+					existingMod.speed = 'Downloaded';
+				} else {
+					vm.downloadingMods = [{ number: modNumber, name: modName, progress: '100', speed: 'Loading' }, ...vm.downloadingMods];
```
roitto: Same thing as above, maybe unshift here? Also Maybe the speed should be `Downloaded` here(?)


```suggestion
					vm.downloadingMods.unshift({ number: modNumber, name: modName, progress: '100', speed: 'Downloaded' });
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+				}
+			});
 		} else {
-			document.getElementById('LoadingStatus').innerText = data.message;
+			if (document.getElementById('LoadingStatus').innerText != data.message) console.log(data.message)
```
roitto: Is it necessary to log here? If so, maybe do it like this for consistency?


```suggestion
			if (document.getElementById('LoadingStatus').innerText != data.message) {
				console.log(data.message);
			}
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
...
+				// Update or add the current mod being downloaded
+				let existingMod = vm.downloadingMods.find(mod => mod.name === modName);
+				if (existingMod) {
+					existingMod.progress = '100';
+					existingMod.speed = 'Downloaded';
+				} else {
+					vm.downloadingMods = [{ number: modNumber, name: modName, progress: '100', speed: 'Loading' }, ...vm.downloadingMods];
+				}
+			});
 		} else {
```
roitto, line 800: Maybe do something like this here for some special micro optimization, so the DOM won't be queried twice? 😀 

```js
const loadingStatusElement = document.getElementById('LoadingStatus`);

// then do the loadingStatusElement.innerText and loadingStatusElement.removeAttribute stuff here..
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -693,16 +695,84 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
 
 	};
 
+	vm.downloadingMods = [];
+	var lastModInfo = ''
```
roitto:
```suggestion
	var lastModInfo = '';
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -630,6 +630,8 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
 		document.getElementById('OriginalLoadingStatus').removeAttribute("hidden");
 		document.getElementById('LoadingStatus').setAttribute("hidden", "hidden");
 		document.getElementById('LoadingServer').style.display = 'none';
+		//vm.downloadingMods = [];
+		lastModInfo = ''
```
roitto:
```suggestion
		lastModInfo = '';
```

ui/modModules/multiplayer/multiplayer.js
```diff
@@ -630,6 +630,8 @@ function($scope, $state, $timeout, $mdDialog, $filter, ConfirmationDialog, toast
 		document.getElementById('OriginalLoadingStatus').removeAttribute("hidden");
 		document.getElementById('LoadingStatus').setAttribute("hidden", "hidden");
 		document.getElementById('LoadingServer').style.display = 'none';
+		//vm.downloadingMods = [];
```
roitto: Why commented out? Should this array be cleared if the popup is closed?

roitto: Also if you want more micro-optimizations `vm.downloadingMods.length = 0;` could work here so there won't be need to allocate a new array. (if angular can handle this :grin: )

ui/modModules/multiplayer/multiplayer.html
```diff
@@ -3,12 +3,35 @@
...
+        <code>{{mod.name}}</code>
+        {{ ' - ' }}
+        <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
+      </span>
+      <div style="flex: 2;" ng-if="mod.progress != 100">
+        <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
+      </div>
+      <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
+        <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
+      </span>
```
roitto: Missing indentation

```suggestion
        <span style="flex: 1;">
          <code>{{mod.name}}</code>
          {{ ' - ' }}
          <span style="color: darkgray;font-size: 0.8em;">{{mod.speed}}</span>
        </span>
        <div style="flex: 2;" ng-if="mod.progress != 100">
          <md-progress-linear md-mode="determinate" value="{{mod.progress}}" class="md-accent"></md-progress-linear>
        </div>
        <span ng-if="mod.progress == 100" style="color: green; font-size: 1em; margin-left: 10px; float: right;">
          <md-icon class="material-icons" style="vertical-align: middle;">check_mark</md-icon>
        </span>
```

ui/modModules/multiplayer/multiplayer.js
Starystars67: Sadly it was reactivity that made me have to do this this way :(

ui/modModules/multiplayer/multiplayer.js
Starystars67: I like the idea of Downloaded but there is a lacking in state for each mod from the launcher currently.
If we could address that I would love to make it Downloaded and then loaded / ready when it is loaded.

Will update it to Downloaded though in the meantime.

ui/modModules/multiplayer/multiplayer.js
Starystars67: Ah good spot thank you!

Starystars67:
image

ui/modModules/multiplayer/multiplayer.html
Starystars67: Strange, the indentation is fine on my side?

TheHawk09:
Looks good! Just the border radius on the progress bar left and it’s all good

ui/modModules/multiplayer/multiplayer.js
Starystars67: I'm not sure on this one, what does everyone else think?

TheHawk09: It’s probably better to capitalize just the main words, that would fit BeamNG’s style more

WiserTixx: There are some messages that are in lowercase. I suppose it could make debugging harder in some cases.

TheHawk09: True, that’s a good point. Maybe just capitalize the main words manually instead of forcing it with CSS. Still, most of them are capitalized anyway.

Starystars67:
image

ui/modModules/multiplayer/multiplayer.html
Starystars67: I see it now, my bad, thank you!

TheHawk09:
> image

Looks good! Just the filled part could use the same rounding treatment as the container, and the colors also seem to be a bit off?

WiserTixx (changes requested):
This PR looks good but I have a few requests.

Could this grey box be hidden when there is nothing to display in it?
image

Same goes for the divider when there is no extra message.

And I've noticed when you quit a server while your downloading and then join another server the old history is partially displayed. Though this may be the launcher's fault.

WiserTixx (approved):

PR #805, opened 2025-09-27
Changed the name of favorite servers when added from direct connect

Olrosse:
Currently if you save a server as favorite from the direct connect tab it will use the IP and Port as the name,
This makes it show the server IP and Port at the top of the screen when in the server, Which makes it very easy to accidentally leak the server IP of a private server
here I've have changed it to use time and date instead

Ideally we would prompt the user to add a name, but i don't have enough UI experience to be able to do that


StanleyDudek:
Could the actual name of the server be picked up from an information packet?

WiserTixx:
The launcher does support this (https://github.com/BeamMP/BeamMP-Launcher/pull/161) but I feel like that's a little out of scope for this PR.

WiserTixx (approved):

PR #813, opened 2025-10-12
Introduce new Community Safety Options

Starystars67:
This PR introduces a simple client side chat filter. This approach was chosen to prevent conflict from aggressive members complaining how they cant use certain words. Instead this works by just replacing matched offensive words according to the word lists with * for each character.

The new settings page looks as follows:

image


lua/ge/extensions/multiplayer/filters/level_3.txt
```diff
@@ -0,0 +1,369 @@
...
+wrapping men
+wrinkled starfish
+wtf
+wttp
+yaoi
+yellow showers
+yiffy
+zoophilia
+Klinkers
+Klinker
```
boubouleuh, line 369: theses aren't censored correctly

boubouleuh (changes requested):
i noticed some issues :
- typing in the custom content filter is laggy for some reasons 
- the chat content filtering is empty before picking a severity
- adding a single letter to a censored word will uncensor it, like "orgasm" to "orgasme" 
- capitalizing letters can uncensor the censored word

PR #817, opened 2025-10-19
Add mod count and total size to security warning dialog

leiflundberg:
Displays the number of mods and total download size in the security warning dialog when joining servers with mods. This helps users make informed decisions about what they're downloading.

The dialog now shows:
- Number of mods
- Total download size (human-readable format)


WiserTixx:
Could you provide some screenshots of this in action?

PR #827, opened 2025-10-21
Fixed rounding of the right corner

5DROR5:


ui/modules/apps/BeamMP-Session/app.js
```diff
@@ -160,14 +161,30 @@ app.controller("Session", ['$scope', '$mdDialog', 'Settings', function ($scope,
...
 	});
 
 	$scope.$on('setPlayerCount', function (event, count) {
 		document.getElementById("Session-PlayerCount").innerHTML = count;
+		updateRightCorners();
 	});
 }]);
 
 
+function updateRightCorners() {
```
roitto: Hmm would it be possible to just apply the border color, radius and overflow:hidden styles to the .outerDiv class in the stylesheet and remove the border rounding styles from the .block classes?

Then this javascript method would not be needed.

5DROR5: To make it work without changing the JS, you need to wrap all the data blocks with a new div in HTML and then change the 2 CSS accordingly.
I preferred to change one file instead of 3 files
If you think this option is better, I can do it

roitto: Imo it's always better to prefer plain html/css to make style related changes if possible.

When using javascript to do dynamic style changes, it's harder to maintain the existing html/css structure because you might break something that is referenced in javascript side and vice versa. (And ofc plain html/css is better if you think about performance.)

I'm not a maintainer of this repo tho. What other people think?

WiserTixx: I'd prefer the more maintainable option.

5DROR5:
Refactor border-radius handling:
moved from JS to HTML and updated CSS accordingly.

5DROR5:
Old:


New:

WiserTixx (changes requested):
The marquee hover effect seemingly doesn't work anymore, and when direct connecting to a server (so the session app just shows players and ping) the queues section doesn't properly round. Also it seems the app isn't centred anymore.

ui/modules/apps/BeamMP-Session/app.css
```diff
@@ -12,7 +12,14 @@
...
 	}
-	.block {
+    	.info-container {
+        display: flex;
+		flex-flow: row nowrap;
+        height: 100%;
+        flex: 1 1 auto;
+        min-width: 0;
+    }
+    	.info-container > .block {
```
roitto: Indentations are bit off here

5DROR5:
> The marquee hover effect seemingly doesn't work anymore, and when direct connecting to a server (so the session app just shows players and ping) the queues section doesn't properly round. Also it seems the app isn't centred anymore.

Could you clarify the problems a bit more, maybe with a screenshot? Everything looks fine on my side.

WiserTixx:
It looks like the hover effect is fine now, I guess I didn't test it properly. Here's screenshots of the rest of the issues.

On development:
image
image

On this branch:
image
image
image
(It's on the center line)

5DROR5:
Sorry, I didn't read carefully that you were talking about a problem when connecting directly to the server. Now I understand what you mean. I'll check it out tomorrow.

ui/modules/apps/BeamMP-Session/app.css
```diff
@@ -12,7 +12,14 @@
 		justify-content: center;
 		flex-flow: row nowrap;
 	}
-	.block {
+    .info-container {
+        display: flex;
+		flex-flow: row nowrap;
+        height: 100%;
+        flex: 1 1 auto;
```
roitto: This is probably causing that weird spacing issue when there are no server name.

Maybe this would work:

```suggestion
        flex: 0 1 auto;
```

Or this:

```suggestion
        flex: 0 0 auto;
```

ui/modules/apps/BeamMP-Session/app.html
```diff
@@ -4,21 +4,22 @@
...
-			<div class="marquee">
-				<span class="session-status"></span>
-				<span class="session-status"></span> 
-				<!-- duplicated for the marquee -->
+		<div class="info-container">
+			<div class="block" id="server-name-block" style="display: none;">
+				<div class="marquee">
+					<span class="session-status"></span>
+					<span class="session-status"></span> 
+					</div>
```
roitto: Indentation

```suggestion
				</div>
```


Also inconsistent indentations in the style file class changes

boubouleuh:
Hi, just a suggestion but i think that opacity on the windows would fit better than a text shadow on every text to improve readability, and it would probably look better

Also, the queue-block background position is constantly updating even when not needed. That could drain a little of performances, i dont know if that have been introduced by this pr though (reworking the animation by an css based one can be good)

5DROR5:
> Hi, just a suggestion but i think that opacity on the windows would fit better than a text shadow on every text to improve readability, and it would probably look better
> 
> Also, the queue-block background position is constantly updating even when not needed. That could drain a little of performances, i dont know if that have been introduced by this pr though (reworking the animation by an css based one can be good)

Hey, thanks for the feedback!

You're right about the queue-block JS, but that's how it was in the original code. I was only focused on the redesign, so I didn't change any of the logic.

As for the text-shadow, I get your point, but this whole redesign came from a private mod I made. Two devs saw it and asked me to contribute this exact look.
I'm also a big fan of how it looks 😉

boubouleuh:
> > Hi, just a suggestion but i think that opacity on the windows would fit better than a text shadow on every text to improve readability, and it would probably look better
> > Also, the queue-block background position is constantly updating even when not needed. That could drain a little of performances, i dont know if that have been introduced by this pr though (reworking the animation by an css based one can be good)
> 
> Hey, thanks for the feedback!
> 
> You're right about the queue-block JS, but that's how it was in the original code. I was only focused on the redesign, so I didn't change any of the logic.
> 
> As for the text-shadow, I get your point, but this whole redesign came from a private mod I made. Two devs saw it and asked me to contribute this exact look. I'm also a big fan of how it looks 😉

I really like the look. There's only the text-shadow that I think doesn't fit really well, not very practical too since you need to disable it for colored text, for example. That's my personal opinion. I don't know if any others think the same, so yeah

WiserTixx (approved):

PR #838, opened 2025-11-22
Garbage cleanup

Olrosse:
I've gone through and tried to clear up a lot of the Lua garbage that gets created, and additionally optimized a few things,

here is what's changed

MPVehicleGE onPreRender loop is cleaned up and now only creates garbage if the nametags distance has changed, this is done by caching nametags, queues, vectors, colors and changing vector math to the garbage free set functions as well as using the XYZ versions of the getPosition/rotation functions.

all be:getObjectByID() calls are switched to the cached getObjectByID() function.

Position packets are not decoded and encoded in GE anymore when sending and receiving, we used it to adjust velocity to game speed when in slow motion, that is now done in VE, received packets are also changed to use mailboxes instead of queues, so we can give the mailbox the position string directly, which saves more string operations,

Receive should now allow 100s of vehicles without garbage stutters, that's assuming the computer can handle that many vehicles ofc xD, sending is still causing quite a few stutters with that many vehicles, but the direct VE sockets Stefan's been working on will fully get rid of that.

Creating packet strings is now done with string buffers, which saves some string operations, it half's the garbage per sent packet.

I've also cleaned up the Imgui chat so it produces significantly less lua garbage, each message is reduced from 320 bytes to 0, so now a filled out imgui chat won't cause slowdowns anymore

### DIFF
--- a/lua/ge/extensions/MPControllerGE.lua
+++ b/lua/ge/extensions/MPControllerGE.lua
@@ -21,7 +21,7 @@ local function sendControllerData(data, gameVehicleID)
 				decodedData.vehID = MPVehicleGE.getServerVehicleID(decodedData.vehID) -- used for controllers that call to another vehicle, like the me262 missile targeting system
 			end
 			data = jsonEncode(decodedData)
-			MPGameNetwork.send(MPHelpers.generatePacketBuffer('Rc',serverVehicleID,data))
+			MPGameNetwork.send(MPNetworkHelpers.generatePacketBuffer('Rc',serverVehicleID,data))
 		end
 	end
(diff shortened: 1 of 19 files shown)


SaltySnail:
We tried testing this, however the positions didn't seem to update correctly. The inputs were syncing correctly though.

Olrosse:
> We tried testing this, however the positions didn't seem to update correctly. The inputs were syncing correctly though.

just joined some servers with this branch and they look like they are working fine, what exactly was not syncing properly?

Starystars67 (approved):

PR #844, opened 2025-12-07
Arcade abs sync

techeyy:
Players using Arcade ABS will no longer appear as if they are locking their tires and creating laggy smoke trails behind them. 

https://www.youtube.com/watch?v=4-NsVA7Zup0


lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -1324,11 +1328,13 @@ local function applyVehSpawn(event)
 		log('W', 'applyVehSpawn', "(spawn)Updating vehicle from server "..vehicleName.." with id "..spawnedVehID)
 		spawn.setVehicleObject(spawnedVeh, {model=vehicleName, config=serialize(vehicleConfig), pos=pos, rot=rot, cling=true})
 		spawnedVeh:setField("protected", 0, protected or "0")
+		spawnedVeh:setField("absMode", 0, absMode or "realistic")
 	else
 		log('W', 'applyVehSpawn', "Spawning new vehicle "..vehicleName.." from server")
 		spawnedVeh = spawn.spawnVehicle(vehicleName, serialize(vehicleConfig), pos, rot, { autoEnterVehicle=false, vehicleName="multiplayerVehicle", cling=true})
 		spawnedVehID = spawnedVeh:getID()
 		spawnedVeh:setField("protected", 0, protected or "0")
+		spawnedVeh:setField("absMode", 0, absMode or "realistic")
```
OfficialLambdax: Unlikely required to set a new field here

lua/ge/extensions/MPVehicleGE.lua
```diff
@@ -1422,6 +1430,7 @@ local function applyVehEdit(serverID, data)
 	end
 	
 	veh:setField("protected", 0, protected or "0")
+	veh:setField("abs", 0, absMode or "realistic")
```
OfficialLambdax: Field is called `absMode` not `abs`

OfficialLambdax (dismissed):
Field set seems unnecessary. Otherwise looks good. Untested

techeyy:
ABS behavior now applied on vehicle reset, no longer "forgets" ABS behavior when CTRL+R is used by the observing player.

Note that when a driver changes their assist setting, proper behavior is only guaranteed for the observer once they queue the driver's change. (non issue)

WiserTixx (approved):

PR #856, opened 2026-01-26
fixed imGui chat resetting settings on launch

Olrosse:
jsonEncode can't encode ImVec4, so now it converts them to a table on save, and back to ImVec4 on load

Missing settings now gets filled in instead of the entire config file being reset

### DIFF
--- a/lua/ge/extensions/UI.lua
+++ b/lua/ge/extensions/UI.lua
@@ -358,7 +358,7 @@ end
 --- This function is used to load the settings and config of the UI (chat)
 local function loadConfig()
     local config = io.open("./settings/BeamMP/chat.json", "r")
-    if not config then -- Write new config
+    if not config or config:read() == "null" then -- Write new config
         log("I", "chat", "No config found, creating default")
 
         local jsonData = jsonEncode(M.defaultSettings)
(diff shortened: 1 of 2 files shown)


lua/ge/extensions/UI.lua
```diff
@@ -358,7 +358,7 @@ end
 --- This function is used to load the settings and config of the UI (chat)
 local function loadConfig()
     local config = io.open("./settings/BeamMP/chat.json", "r")
-    if not config then -- Write new config
+    if not config or config:read() == "null" then -- Write new config
```
boubouleuh, line 361: Can the config file really only contain "null" in it?

lua/ge/extensions/UI.lua
```diff
@@ -398,15 +396,31 @@ local function loadConfig()
...
     configLoaded = true
-
-    if #findMissingKeys(M.defaultSettings, settings) > 0 then
-        log('I', "BeamMP", "Missing one or more settings, resetting config file...")
-        M.settings = deepcopy(M.defaultSettings)
-        optionsWindow.saveConfig(M.settings) -- we pass it in because "UI.lua" and "ui/options.lua" depend on eachother,
-                                             -- so instead of doing "UI.options", we pass it in instead.
-        return
+    local missingVars = findMissingKeys(M.defaultSettings, settings)
+    dump(missingVars)
```
boubouleuh: that dump can be removed

boubouleuh (changes requested):

lua/ge/extensions/UI.lua
Olrosse: probably not, it did when i was working on this because i broke the write at some point, which broke the chat loading entirely, So i figured it didn't hurt to leave it in, But i don't think it's been an issue so i could probably remove it again

boubouleuh: seems oka to leave it then i guess

boubouleuh (approved):

WiserTixx (approved):

PR #893, opened 2026-06-22
Switched to batch based mod loading

### LINKED ISSUE #803: [Bug] Large mod folder on server causes clients to crash on load
Yes

**Describe the bug**
When the size of the mod folder on the BeamMP server approaches ~10.8-11GB, attempting to join the server will cause the game to crash with an 0x00000001 error every time, even if the game will load into single player using the exact same loadout of mods. The crash logs left behind by the BeamNG client when this occurs have no useful information/contain no C++ trace. [beamng.1.log](https://github.com/user-attachments/files/22537842/beamng.1.log)

**To Reproduce**
1. Add at least ~10.8GB of mods into the Resources/Client/ directory on the BeamMP server
2. Start the server
3. Attempt to join the server from the game
4. Wait for game to download any missing mods
5. Game crashes and error is thrown before getting to the loading splash screen

**Expected behavior**
Game proceeds to loading splash screen and then loads into the map as normal

**Screenshots**

 - OS + Version: Windows 11 x64 24H2

**Additional context**
Issue affects all players that join my server regardless of OS or game version, issue has persisted across multiple patches to both BeamNG and BeamMP.

WiserTixx:
Instead of everything at once

Derived from @coops924 (https://github.com/BeamMP/BeamMP/issues/803#issuecomment-4435219031)

Seemingly fixes a crash that can occur when BeamNG is overloaded by mod updates. Should fix #803

### DIFF
--- a/lua/ge/extensions/MPModManager.lua
+++ b/lua/ge/extensions/MPModManager.lua
@@ -148,13 +148,22 @@ end
 --- Load the Servers mods, these are put in by the BeamMP Launcher
 local function loadServerMods()
 	log('W', 'loadServerMods', 'loadServerMods')
-	
+
 	local modsDir = FS:findFiles("/mods/multiplayer", "*.zip", -1, false, false)
-	for _, modPath in pairs(modsDir) do
-		core_modmanager.workOffChangedMod(modPath, 'added')
-	end
-	checkAllMods()
-	MPCoreNetwork.requestMap()
+
+    core_jobsystem.create(function(job)
+		local amount = #modsDir
+        local batchSize = 5
+		local core_modmanager_workOffChangedMod = core_modmanager.workOffChangedMod
+        for i = 1, amount, batchSize do
+            for j = i, math.min(i + batchSize - 1, amount) do
+                core_modmanager_workOffChangedMod(modsDir[j], 'added')
+            end
+            job.sleep(0)
+        end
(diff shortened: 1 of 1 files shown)


lua/ge/extensions/MPModManager.lua
```diff
@@ -148,13 +148,22 @@ end
...
-	for _, modPath in pairs(modsDir) do
-		core_modmanager.workOffChangedMod(modPath, 'added')
-	end
-	checkAllMods()
-	MPCoreNetwork.requestMap()
+
+    core_jobsystem.create(function(job)
+		local amount = #modsDir
+        local batchSize = 5
+		local core_modmanager_workOffChangedMod = core_modmanager.workOffChangedMod
```
OfficialLambdax, line 157: Comparing the suggested change and this code piece. Why do you cache a reference to this? Is there probably cause to believe/prevent the core_manager extension goes missing or manipulated during load?

WiserTixx: I don't think anything happens to the function or extension (we would've most likely noticed that by now) It was just a small optimization because the function is basically spam called

OfficialLambdax: Saves a tiny bit of a table lookup time true

OfficialLambdax (approved):
Looks good to me!

PR #895, opened 2026-06-23
fixed partial receive

Olrosse:
fixed partial receive applying the wrong data
removed debug dumps

this should fix the server list decode error


Starystars67 (approved):

SaltySnail (approved):
LGTM

PR #896, opened 2026-07-05
Guard nil partials in MPNetworkHelpers.receive + bound the sendData retry loop

IrPgFKS0:
### Status update — narrowed

Half of item 1 has since landed on `development` independently: `MPNetworkHelpers.receive` now uses `fullBody` on the completing read, so the packet-tail truncation this PR originally led with is **already fixed upstream**. I re-verified the remainder against current `development`; two things are still present, and this PR is now just those.

### Problem

**1. `partialBody` is used unguarded where it can be `nil`.**

On a pure 0-byte timeout the partial comes back `nil`, and the body path then does:

```lua
recvState.data = partialBody
recvState.missing = len - #partialBody          -- #nil -> error on the GE thread
...
recvState.missing = recvState.missing - #partialBody
```

The *header* path four lines above already guards exactly this case (`partialHeader == nil or partialHeader == ""`); the body path doesn't. Three sites, one `or ""` each.

**2. `MPGameNetwork.sendData` can freeze the whole game.**

```lua
local retries = 1
...
while (retries > 0 and error) do
```

`retries` is never decremented, so the bound is a no-op — a persistent timeout (launcher stalled, send buffer full) spins forever **on the GE thread**, which is the scenario the function's own `TODO` warns about. If a partial send makes no progress, `index` stays `0` and the same bytes are re-sent indefinitely.

### Fix

- Guard the three `partialBody` uses.
- Decrement the retry counter so the loop is bounded as intended, and stop retrying on non-timeout errors (e.g. `closed`) — the existing error handler below already deals with those.

A note on scope: the fork this came from goes further on (2), retrying only while the send makes *progress* (`index` advances) and giving up after a 0.25 s stall window, because abandoning a partially-sent packet corrupts the length-prefixed stream. That's a bigger change than this PR proposes; the minimal bound above is what's here. Happy to swap in the progress-based version if you'd prefer it.

### How this was found

Both were found during a robustness audit of a LAN fork of BeamMP and have been running there for months of sessions.

### Transparency

These fixes come from an AI-assisted fork: the bugs were found and the patches written with the help of an AI coding tool (Claude), then tested by a human in real multiplayer sessions. Given this project's policy on AI-generated code, this is submitted as a **draft** for the maintainers to decide — happy to close it if that's not wanted, or for a maintainer to re-implement independently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lua/ge/extensions/MPGameNetwork.lua
+++ b/lua/ge/extensions/MPGameNetwork.lua
@@ -96,7 +96,11 @@ local function sendData(data) -- TODO currently the socket keeps retrying indefi
 				packet = string.sub(packet, index + 1)
 
 				bytes, error, index = TCPLauncherSocket:send(packet)
+			else
+				break -- non-timeout error (e.g. closed): stop retrying, the handler below deals with it
 			end
+			retries = retries - 1 -- was never decremented: a persistent timeout (frozen/stalled
+			                      -- launcher) spun this loop forever on the GE thread = game freeze
 		end
 	end
 
--- a/lua/ge/extensions/MPNetworkHelpers.lua
+++ b/lua/ge/extensions/MPNetworkHelpers.lua
@@ -41,12 +41,12 @@ M.receive = function(launcherSocket, recvState)
 		local fullBody, recvStatus, partialBody = launcherSocket:receive(len)
 		if not fullBody then
 			if recvStatus == 'timeout' then
-				-- combine the previously received partial (if any) with this partial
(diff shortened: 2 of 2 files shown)


IrPgFKS0:
Narrowed this description: the packet-tail fix it originally led with has since landed on `development` independently, so what's left is the `nil` partial guards (the header path already guards that case, the body path doesn't) and the `retries` counter that's never decremented. Re-verified both against current `development` today.

PR #900, opened 2026-07-29
Ash's attempt to do development

ashmaker000:


lua/ge/extensions/core/cameraModes/passenger.lua
Starystars67, line 1: I think these camera names/hooks are game default?
Can you double check this.

lua/ge/extensions/MPCoreNetwork.lua
Starystars67, line 1: fine with the rename if its to prevent conflict but can we stick to the onXXXX format so we know it is triggered on that event rather than a generic calling or something.

For example onBeamMPAuthRecevied would be fine, or even onBeamMPLauncherAuthenticated.

Same applies to all the vents here.

lua/ge/extensions/MPModManager.lua
```diff
@@ -28,39 +29,21 @@ local function unloadLocalesAndDefaults()
 	FS:directoryRemove('/temp/beammp')
 end
 
---- Load the BeamMP provided locales + defalts and merge them into a working set for BeamNG
-local function loadLocalesAndDefaults() -- loads beammp locales and default settings without having to directly replace the game locales and default settings
+--- Load the BeamMP defaults and merge them into a working set for BeamNG.
+--- BeamNG 0.39 loads mod translations directly from /locales/translations.
+local function loadLocalesAndDefaults()
```
Starystars67, line 34: I think we can remove this as a whole for the locales now.

Should keep the defaults but I think it also needs fixing possible as I see warnings in the console still.

lua/ge/extensions/MPModManager.lua
```diff
@@ -213,10 +202,21 @@ local function setServerMods(modsString)
...
 	end
 	log('M', 'setServerMods', 'Server Mods set to: ' .. dumps(mods))
-	for key, modName in pairs(mods) do -- mods in a directory deeper than /mods/ have "<directory name> + modname" as their mod name
+	serverModDisplayNames = {}
+	for key, modName in ipairs(mods) do -- mods in a directory deeper than /mods/ have "<directory name> + modname" as their mod name
+		serverModDisplayNames[key] = modName
 		mods[key] = string.lower('multiplayer'..modName)
 	end
 	serverMods = mods
+	guihooks.trigger("BeamMP_ServerModsChanged", serverModDisplayNames)
```
Starystars67, line 211: Again lets look at doing the onBeamMPXXX detail.

ui/modModulesOLD/multiplayerPause/multiplayerPause.js
Starystars67, line 1: Not required since thiss will be deleted

ui/modules/apps/BeamMP-Chat/app.html
Starystars67, line 1: Lets review what we need to keep here from the OLD modules folder and what we can condense to the new one.

ui/ui-vue/mods/BeamMP/index.js
```diff
@@ -20,7 +20,7 @@ const TITLE_translationId = 'ui.beammp.common.title'
 function addBeamMPMainMenuButton(addButton) {
   addButton({
     // Old Angular-compatible menu button shape.
-    translateId: TITLE_translationId,
+    translateid: TITLE_translationId,
```
Starystars67, line 23: This is not needed if we are using the peopleOutline since it will use the vue setup rather than the angular way of working.

Need to work out if we are using the account-multiple one below or the peopleOutline.

Starystars67 (changes requested):

lua/ge/extensions/core/cameraModes/passenger.lua
ashmaker000: Will fix in next commit

lua/ge/extensions/MPCoreNetwork.lua
ashmaker000: Will fix in next commit

lua/ge/extensions/MPModManager.lua
ashmaker000: Will fix in next commit

lua/ge/extensions/MPModManager.lua
ashmaker000: Will fix in next commit

PR #936, opened 2026-08-01
fixed ui being set to the wrong layout

### LINKED ISSUE #933: Exiting the Radial Menu doesn't return to the BeamMP HUD layout
To reproduce:
1. Join a server
2. Be on the BeamMP HUD layout
3. Open and close the radial menu

Olrosse:
should fix the missing ui elements and no ui when closing the radial menu

setGameState was set in two different places, and one was setting it to the old layout, i've just removed the one that was setting it wrong,
the correct on is located in runPostJoin in MPCoreNetwork

### DIFF
--- a/lua/ge/extensions/beammp/multiplayer.lua
+++ b/lua/ge/extensions/beammp/multiplayer.lua
@@ -141,7 +141,6 @@ local function onWorldReadyState(state)
 	if state == 2 then
 		if MPCoreNetwork and MPCoreNetwork.isMPSession() then
 			log('M', 'onWorldReadyState', 'Setting game state to BeamMP multiplayer.')
-			core_gamestate.setGameState('multiplayer', 'multiplayer', 'multiplayer')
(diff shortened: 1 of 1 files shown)


lua/ge/extensions/beammp/multiplayer.lua
```diff
@@ -141,7 +141,6 @@ local function onWorldReadyState(state)
 	if state == 2 then
 		if MPCoreNetwork and MPCoreNetwork.isMPSession() then
 			log('M', 'onWorldReadyState', 'Setting game state to BeamMP multiplayer.')
-			core_gamestate.setGameState('multiplayer', 'multiplayer', 'multiplayer')
```
Starystars67, line 144: Hey Olrosse, I believe I updated the latest to have the 2nd param here set to beammp which then should use the beammp layout. I notice this is not the case in this PR and also that your removing it.

It was done so it did not conflict with the base game, whats the logic?

Starystars67: I should learn to read better :(
Sorry

Starystars67 (approved):

PR #952, opened 2026-08-03
GC optimization for positionVE and velocityVE

stefan750:
Reduces the garbage generated by positionVE by about 75% from an average of ~4000 bytes to less than 1000 bytes per run of updateGFX.
Performance timings of the functions themselves are mostly the same, maybe a marginal improvement. It should help reduce stutters caused by garbage collection though.

Changes made:
- Updated most vector maths to use garbage free set...() functions or StackVec3
- Use cached variables for vec3s instead of creating new ones where possible
- Switched to XYZ variants of vehicle functions to avoid creating new vec3s there
- Updated sending code to reuse the table instead of recreating it and use a StringBuffer to assemble the string for queueing to GE

Initial testing watching people drive around in public servers seems fine.

### DIFF
--- a/lua/vehicle/extensions/BeamMP/positionVE.lua
+++ b/lua/vehicle/extensions/BeamMP/positionVE.lua
@@ -9,6 +9,7 @@ local M = {}
 local abs = math.abs
 local min = math.min
 local max = math.max
+local sqrt = math.sqrt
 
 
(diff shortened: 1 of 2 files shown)


Olrosse (approved):
seems to work as expected

PR #974, opened 2026-09-14
Ignore stale position mailbox data when a server vehicle ID is reused

foodcache3:
## Problem

A remote vehicle can spawn at the position of the **previous** vehicle that used the same server vehicle ID and then ignore all real position packets for a while. It is easiest to hit with walking mode: every get-out spawns a new unicycle that reuses the previous unicycle's ID, so another player sees your beamling in the wrong place, sometimes for a long time.

## Cause

`positionGE.applyPos` delivers packets through the mailbox `"vehPosPckt" .. serverVehicleID`. When a vehicle is removed, that mailbox keeps its last packet, and the server hands the freed ID to the owner's next vehicle.

The new vehicle's `positionVE.updateRemoteData` starts with `lastMailboxVersion = 0`, so the leftover packet looks unread:

1. It is applied, and the forced teleport on frame 6 (`framesSinceReset == 6`) moves the new copy to the old vehicle's last position.
2. `remoteData.timer` takes the old packet's `tim`. `tim` is the sender's per-vehicle timer and restarts at 0 for the new vehicle, so every real packet fails `if remoteData.timer > tim then return end`. They keep being rejected until the new vehicle has existed as long as the old one did.

It is a race. It only happens when the new vehicle's first real packet reaches the mailbox after its VE first reads it, which is why it shows up inconsistently depending on load and timing.

Captured with temporary logging on the receiving client: the new unicycle copy sat exactly on the previous unicycle's last position, 5 to 7 m from every incoming packet, with packets arriving at about 60/s. It stayed there for about 3 s, roughly the previous unicycle's age, then snapped to the correct position.

## Fix

In `updateRemoteData`, the first time a vehicle sees its mailbox name (or whenever the name changes), it records the mailbox's current version as already read. Anything left from a previous vehicle is ignored, and the next real packet is processed normally. Skipping that first read costs at most one packet (one frame of data).

## Testing

Tested with two clients on one server, logging on player A each remote unicycle's actual position against the latest received packet, once per second. The repro was done on a server setup where the race reproduces reliably.

- **Before** (current `development`): player B walks around, gets into a vehicle, gets straight back out. On player A's screen, B's new beamling appears at the previous beamling's last spot, several meters from the incoming packets, while packets keep arriving at about 60/s. It doesn't follow B until the stale timestamp is passed.
- **After** (this patch, same setup and routine): B's new beamling appears at the right spot and follows B from the first packet on, across repeated get in/out cycles.


Olrosse (dismissed):
looks good,

I'm unable to reproduce the bug somehow, but it makes sense it would happen, i assume it's a race condition if a packet arrives or not to update the mailbox before the mailbox is first read, i should have thought of that when i switched it to mailboxes,

lua/vehicle/extensions/BeamMP/positionVE.lua
```diff
@@ -224,6 +225,15 @@ local function updateRemoteData()
 	if not v.mpServerID or v.mpServerID == "" then return end
 	local mailBoxName = "vehPosPckt" .. v.mpServerID
 	local currentMailBoxVersion = obj:getLastMailboxVersion(mailBoxName)
+	-- Server vehicle IDs get reused (e.g. every unicycle), and the mailbox still holds the last packet
+	-- of the previous vehicle with that ID. Treat whatever is already there as read, otherwise this
+	-- vehicle teleports to the old vehicle's last position and rejects every real packet until its
+	-- sender timer passes the old vehicle's timestamp.
```
WiserTixx: Could this comment be removed?

WiserTixx (changes requested):

WiserTixx:
Also could your branch be rebased?

foodcache3:
Just rebased and removed the comment. Should be good to go.

WiserTixx (approved):

PR #975, opened 2026-09-15
Add role nil check

WiserTixx:


Starystars67 (approved):

Starystars67:
Looks good to me but do we need to take into account the new primary value in the returned login data for the chosen role to show over the collection of roles that a user might have?

WiserTixx:
Why does the api not return the primary role in `role`?

Starystars67:
Because you might have multiple roles that grant extra or other things.

PR #979, opened 2026-09-18
Add/player profile details roles + banner patreon

bptisteg:
# Player Profile & Patreon Banner 

This PR adds several new features including the player profile in the serverlist menu, copy + see your ID, and an patreon banner that adapts to player status EA

### Profile Section

- Display player account information with role details, and you can click on your id to copy it.
userprofile
image

### Patreon Banner

- The previously empty space has been converted to a message which changes if you are a donator, and the read more button will redirect you to the patreon page

### Small changes
- Edited menu button for better integration
menu

- Switched position between player stats and banner, banner are centered compared to screen

ea
user


Starystars67:
Looking good visually there!

One thing I notice, it still have the green Patreon notice even if your EA in the left sidebar. Might be nice to change that for a thank you in purple instead?

Also rather than saying the role as EA in the top right, I'd recommend using the expanded name :)

bptisteg:
the lua code didn't allow for it, but we had a really cool guy named Tix who helped me add support for extended names..and NOW that's work!!
roleprofile

### Leftbar patreon
patreonleftbar

### Guest account
I hadn't thought of that (😭) Tix (again) reminded me, and I added profile support for guest accounts
guestprofile

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+  const role = state.auth.value?.role || ""
+  return role === "EA" || role === "Early Access"
+})
+
+const isGuest = computed(() => {
+  const role = state.auth.value?.role || ""
+  const username = state.auth.value?.username || ""
+  return role === "GUEST" || role === "Guest" || role === "guest" ||
+         username.toLowerCase().startsWith("guest")
+})
```
WiserTixx: instead of doing this can we not just hide the id text if there is no id? people can have names starting with "guest"

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+const accountId = computed(() => {
+  // Try to get the ID from various possible fields
+  return state.auth.value?.id || state.auth.value?.Key || state.auth.value?.identifiers?.beammp || "N/A"
+})
+const accountRoleColor = computed(() => {
+  return state.auth.value?.color || "transparent"
+})
+const isSpecialRole = computed(() => {
+  const role = state.auth.value?.role || "USER"
+  return role && role !== "USER" && role !== "User" && role !== "user"
```
WiserTixx: Only USER is valid. Also the color value will be null if the role is user so that can be checked directly in the template. The other usage of this function can also just check if role !== "USER" directly

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+const accountRole = computed(() => {
+  const roleTag = state.auth.value?.roleInfo?.tag || ""
+  return roleTag.replace(/^\s*\[|\]\s*$/g, "").trim() || "USER"
+})
+const accountId = computed(() => {
+  // Try to get the ID from various possible fields
+  return state.auth.value?.id || state.auth.value?.Key || state.auth.value?.identifiers?.beammp || "N/A"
+})
+const accountRoleColor = computed(() => {
+  return state.auth.value?.color || "transparent"
```
WiserTixx: when this is displayed there will also be a value for color, no need to default to transparent

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
- 
   return state.auth.value?.username
 })
-const accountRole = computed(() => state.auth.value?.role || "User")
+const accountRole = computed(() => {
+  const roleTag = state.auth.value?.roleInfo?.tag || ""
+  return roleTag.replace(/^\s*\[|\]\s*$/g, "").trim() || "USER"
+})
+const accountId = computed(() => {
+  // Try to get the ID from various possible fields
```
WiserTixx: Was this comment meant to be left in?

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+  return state.auth.value?.id || state.auth.value?.Key || state.auth.value?.identifiers?.beammp || "N/A"
```
WiserTixx: authentication will always return an id for non-guest users, not sure where you got the other fields from. "N/A" isn't necessary, it will only show for guests and in that case the field shouldn't be computed

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+  return state.auth.value?.color || "transparent"
+})
+const isSpecialRole = computed(() => {
+  const role = state.auth.value?.role || "USER"
+  return role && role !== "USER" && role !== "User" && role !== "user"
+})
+
+const isEARole = computed(() => {
+  const role = state.auth.value?.role || ""
+  return role === "EA" || role === "Early Access"
```
WiserTixx: Only EA is valid. It would also be nice if ET, MDEV and STAFF were also shown the thank you for supporting text.

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
+const isGuest = computed(() => {
+  const role = state.auth.value?.role || ""
+  const username = state.auth.value?.username || ""
+  return role === "GUEST" || role === "Guest" || role === "guest" ||
+         username.toLowerCase().startsWith("guest")
+})
 
 function useFallbackAvatar(event) {
-  const image = event.currentTarget
+  const image = event.currentTarget
```
WiserTixx: Could the added trailing space be removed?

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
 
+function copyAccountId() {
+  const id = accountId.value
+  if (id && id !== "N/A") {
+    navigator.clipboard.writeText(id)
+      .then(() => {
+        console.log("ID copied to clipboard:", id)
+      })
+      .catch(err => {
+        console.error("Failed to copy ID:", err)
```
WiserTixx: Could this be paired with a visual ui toastr notification like in the old version of this ui? There should be an example of this in the direct connect favourite code

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -15,30 +15,47 @@
...
+      </div>
+
+      <div class="patreon-banner" :class="{ 'patreon-banner--ea': isEARole }" @click="openExternal('https://www.patreon.com/BeamMP')">
+        <div class="patreon-content">
+          <img src="/ui/assets/BeamMP/icons/PATREON_SYMBOL_1_WHITE_RGB.svg" alt="Patreon" class="patreon-icon" />
+          <div class="patreon-text">
+            <span class="patreon-message" :style="{ color: isEARole ? '#ff69b4' : '' }">{{ isEARole ? $tt("ui.beammp.patreon.message.ea") : $tt("ui.beammp.patreon.message.user") }}</span>
+          </div>
         </div>
+        <BngButton class="patreon-button" :accent="ACCENTS.secondary" :style="{ visibility: isEARole ? 'hidden' : 'visible' }">
```
WiserTixx: why is accent a : value? wouldn't accent="secondary" work here?

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -15,30 +15,47 @@
...
           <div class="account-details">
-            <strong>{{ accountName }}</strong>
-            <small>{{ accountRole }}</small>
-          </div>
-          <BngButton accent="secondary" class="logout-button" @click="handleLogout">
+            <small v-if="isSpecialRole" class="account-role-badge">{{ accountRole }}</small> 
+            <div class="account-name-wrapper">
+              <strong>{{ accountName }}</strong>
+            </div>
+            <div v-if="!isGuest" class="account-id-wrapper" @click="copyAccountId" :title="$tt('ui.beammp.accounts.copyid')">
```
WiserTixx: instead this could then check if the ID isn't null and then display it. see the comments for more context

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
   return avatar
 })
 const accountName = computed(() => {
- 
   return state.auth.value?.username
 })
-const accountRole = computed(() => state.auth.value?.role || "User")
+const accountRole = computed(() => {
+  const roleTag = state.auth.value?.roleInfo?.tag || ""
+  return roleTag.replace(/^\s*\[|\]\s*$/g, "").trim() || "USER"
```
WiserTixx: Why default to user here instead of the role code? I feel like this would make more sense and it would still show the role (and subsequently that something isn't right)

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +238,55 @@ const accountAvatar = computed(() => {
...
 function useFallbackAvatar(event) {
-  const image = event.currentTarget
+  const image = event.currentTarget 
   image.onerror = null
   image.src = fallbackAvatar
 }
 
+function copyAccountId() {
+  const id = accountId.value
+  if (id && id !== "N/A") {
```
WiserTixx: here it could use the id from the state auth value directly, it would also remove the need for the "N/A" check

WiserTixx (changes requested):

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: Yes, I left it intentionally to clarify the purpose of the fallback logic.. happy to remove it if you think it's unnecessary

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: indeed, I was using an older approach

changed to : `<BngButton class="patreon-button" accent="secondary"`

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: I included the case insensitive checks initially for robustness in case the role naming conventions changed in the future.. who know 😅

changed to `  return role !== "USER"`

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: ah right, I thought a backend change was needed to avoid including the entire staff list, but that works too

changed to `  return role === "EA" || role === "ET" || role === "MDEV" || role === "STAFF"`

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: make sense 

changed to ` const id = state.auth.value?.id`

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: I use the old code provided it now uses `bngApi`

Image

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -159,6 +180,7 @@ const { events } = useBridge()
 const route = useRoute()
 const router = useRouter()
 const bngVue = window.bngVue || { goBack() {} }
+const bngApi = window.bngApi
```
WiserTixx: isn't bngApi a global?

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
```diff
@@ -217,17 +239,42 @@ const accountAvatar = computed(() => {
...
 function useFallbackAvatar(event) {
   const image = event.currentTarget
   image.onerror = null
   image.src = fallbackAvatar
 }
 
+function copyAccountId() {
+  const id = state.auth.value?.id
+  if (id) {
+    bngApi.engineLua(`setClipboard("`+id+`")`)
```
WiserTixx: There is a possibility of code injection here

WiserTixx (changes requested):

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: I wasn’t sure whether to add them, so I decided to add..

Removed 👍

ui/ui-vue/mods/BeamMP/layouts/BeamMPMain.vue
bptisteg: Im not sure what you mean, can you please elaborate ?

WiserTixx (approved):

